PR #151, opened 2023-07-26
Update HotShot

jbearer:
This is a big change as it not only updates to async HotShot from the run_view_refactor branch, it is also the first time we have run the query service tests with sequencing consensus.

Actually, most of the non-trivial changes stem from the latter. In particular, the indexes by block hash and transaction hash have been updated to only return the earliest instance in case of duplicates. With validating consensus, we never had duplicates. The mock types and network for testing have also been reworked significantly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2867,7 +2867,7 @@ dependencies = [
 
 [[package]]
 name = "hotshot-query-service"
-version = "0.0.6"
+version = "0.0.7"
 dependencies = [
  "async-compatibility-layer",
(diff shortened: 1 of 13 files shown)


shenkeyao (approved):

jbearer:
Converting this to a draft. I will merge it once I'm sure the sequencer works with this change, but it is lower priority now that Noah has fixed the bug with the paper_benchmarking branch

nomaxg (approved):
LGTM

PR #221, opened 2023-09-13
Update to latest HotShot version (tag 0.4.14)

nomaxg:
Key changes
- Switch from ```Ed25519``` to ```BN254```
- Added stake table related fields (known_nodes_with_stake, verification key)
- Switch back to ```submit_transaction``` (from ```send_transaction```)


elliedavidson (approved):
Looks great!

PR #245, opened 2023-11-08
Feat/generalize data source trait

### LINKED ISSUE #242: Make `AvailabilityDataSource` more abstract
Currently, we assume in the `AvailabilityDataSource` interface that blocks are always looked up by height. If you want to look up a block by something else, e.g. hash, you call `get_block_index_by_hash` to get the height, and then `get_nth_block_iter` to get the block. These interfaces are efficiently implemented by the `AppendLog` backend. However, this is a very inefficient way to use a potential SQL backend: looking up a block by hash results in two separate queries to the database, when SQL indices would easily allow us to look up the block itself by hash efficiently. Moreover, the iterator interface is problematic, because in the SQL backend it will result in either a separate query each time `next()` is called, or a single query which loads many blocks, even though not all of the blocks might be consumed. This iterator interface is also never actually used except to load a single block.

We should change the interfaces provided by this trait to be more representative of the intent (e.g. "load a block by hash") and let the specific backend figure out how best to implement this intent.

jbearer:
Adapt the data source traits to be more amenable to the new SQL implementation:
* Make the trait operations higher level, so that the API layer can request a complex operation and the SQL backend can accomplish it in a single query, instead of breaking the operation down into multiple intermediate queries
* Make the trait functions async, so we can avoid blocking executor threads on I/O. This will be especially important when we add the network latency of a round-trip to a remote SQL database.

Closes #242

### DIFF
--- a/src/availability/data_source.rs
+++ b/src/availability/data_source.rs
@@ -11,11 +11,13 @@
 // see <https://www.gnu.org/licenses/>.
 
 use super::query_data::{
-    BlockHash, BlockQueryData, LeafHash, LeafQueryData, QueryableBlock, TransactionHash,
-    TransactionIndex,
+    BlockQueryData, LeafQueryData, QueryableBlock, TransactionHash, TransactionIndex,
 };
-use crate::{Block, Deltas, Resolvable};
+use crate::{Block, Deltas, Leaf, Resolvable};
 use async_trait::async_trait;
+use commit::{Commitment, Committable};
+use derivative::Derivative;
+use derive_more::{Display, From};
 use futures::stream::Stream;
 use hotshot_types::traits::{
     node_implementation::{NodeImplementation, NodeType},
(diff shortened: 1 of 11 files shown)


ImJeremyHe (approved):

src/availability/data_source.rs
```diff
@@ -15,51 +15,105 @@ use super::query_data::{
...
+}
+
+impl<H: Display> Display for ResourceId<H> {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        match self {
+            Self::Number(n) => write!(f, "{n}"),
+            Self::Hash(h) => write!(f, "{h}"),
+        }
+    }
+}
```
sveitser, line 43: What do you think about adding 

```
impl<H> From<usize> for ResourceId<H> {
    fn from(n: usize) -> Self {
        Self::Number(n)
    }
}
```

to reduce a bit of boilerplate?

sveitser (approved):
Looks like a great improvement.

src/availability/data_source.rs
```diff
@@ -15,51 +15,105 @@ use super::query_data::{
...
+use snafu::Snafu;
 use std::error::Error;
-use std::fmt::Debug;
+use std::fmt::{self, Debug, Display, Formatter};
+use std::ops::RangeBounds;
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum ResourceId<H> {
+    Number(usize),
+    Hash(H),
```
nomaxg, line 33: Does it make sense to constrain this as the Id of something Committable?

src/availability/data_source.rs
jbearer: Good suggestion, will do

src/data_source.rs
```diff
@@ -260,70 +265,163 @@ where
     }
 }
 
-#[derive(Clone, Copy, Debug, Snafu)]
-#[snafu(display("unable to open stream"))]
-pub struct StreamError;
+async fn range_stream<T>(
+    mut iter: Iter<'_, T>,
+    range: impl RangeBounds<usize>,
```
nomaxg, line 270: Why make this an `impl RangeBounds<usize>` instead of a `Range<usize>`?

nomaxg (approved):
LGTM just a couple of questions

src/availability/data_source.rs
nomaxg: ahh yeah that's annoying, ok sounds good

src/data_source.rs
jbearer: Because `from..to` is a `Range`, but `..to` is a `RangeTo`, and there are also `RangeFrom` and some others. So the way to deal with all types of ranges is to take `RangeBounds`. It's kind of annoying, I sort of think they should all be the same type of struct with optional lower and upper bounds, but that's how std does it

src/availability/data_source.rs
jbearer: I thought about this more and I kind of convinced myself to like it. We can do `ResourceId<T: Committable>` with `Hash(Commitment<T>)`, which enforces the semantics we want, and I just think `ResourceId<Block>` makes more sense to read than `ResourceId<BlockHash>`. Plus I can also do `From<Commitment<T>> for ResourceId<T>`. I think this is a nice enough ergonomics improvement to justify adding `T: Committable` in a few more places

PR #249, opened 2023-11-10
Feat/sql data source

jbearer:
Draft PR so I can test new CI workflow


src/data_source/versioned_channel.rs
```diff
@@ -0,0 +1,86 @@
...
+                }
+            }
+        })
+    }
+
+    /// Tentatively send a message to the channel.
+    ///
+    /// The message is not sent immediately, but will be delivered to receivers after
+    /// [`commit`](Self::commit) is called.
+    pub(super) fn send(&mut self, msg: T) {
```
sveitser, line 63: Wonder if we could call it something else than `send` to make it more explicit that the values are not sent (yet). I wonder if there's a better name for this type than `VersionedChannel` because it seems like the main features are buffering messages until flushing is desired and reverting messages that haven't been flushed yet but there's nothing like version numbers or jumping around between different versions. For context: when I first jumped to the file I was expecting a channel that can be upgraded.

src/data_source/sql.rs
```diff
@@ -586,20 +622,25 @@ impl Drop for SqlDataSource {
...
     /// Atomically commit to all outstanding modifications to the data.
     ///
     /// If this method fails, outstanding changes are left unmodified. The caller may opt to retry
     /// or to erase outstanding changes with [`revert`](Self::revert).
     async fn commit(&mut self) -> Result<(), Self::Error> {
-        if !self.tx_in_progress {
-            // Nothing to do if there is no outstanding transaction.
-            return Ok(());
+        if self.tx_in_progress {
+            self.client.batch_execute("COMMIT").await?;
```
sveitser, line 639: There are currently no tests failing for me if I remove the two lines 639-40.

jbearer: How did you even notice this!

This would be because all the tests use the same SQL connection for all queries, so the queries are all executing against the same uncommitted transaction. I will add a test that the results of some updates are visible even if creating an entirely new database connection

src/data_source/sql.rs
```diff
@@ -608,134 +649,408 @@ impl VersionedDataSource for SqlDataSource {
...
+    async fn get_block<ID>(&self, id: ID) -> QueryResult<BlockQueryData<Types>>
     where
         ID: Into<BlockId<Types>> + Send + Sync,
     {
-        todo!()
+        let (where_clause, param): (&str, Box<dyn ToSql + Send + Sync>) = match id.into() {
+            ResourceId::Number(n) => ("h.height = $1", Box::new(n as i64)),
+            ResourceId::Hash(h) => ("h.hash = $1", Box::new(h.to_string())),
+        };
+        // ORDER BY h.height ASC ensures that if there are duplicate blocks, we return the first one.
```
sveitser, line 714: I know there are multiple comments in the code about this but I couldn't find an open issue: 
- https://github.com/EspressoSystems/hotshot-query-service/issues/284

src/data_source/sql.rs
```diff
@@ -608,134 +649,408 @@ impl VersionedDataSource for SqlDataSource {
...
     }
 
-    async fn get_leaf_range<R>(&self, _range: R) -> QueryResult<Self::LeafRange<'_, R>>
+    async fn get_leaf_range<R>(&self, range: R) -> QueryResult<Self::LeafRange<'_, R>>
     where
         R: RangeBounds<usize> + Send,
     {
-        todo!()
+        let (where_clause, params) = bounds_to_where_clause(range, "height");
+        let query = format!("SELECT leaf, qc FROM leaf{where_clause} ORDER BY height ASC");
```
sveitser, line 739: Nit: looks a bit weird without the space between `leaf{where_clause}`.

src/data_source/sql.rs
jbearer: Thank you

src/data_source/versioned_channel.rs
jbearer: Done in [514d444](https://github.com/EspressoSystems/hotshot-query-service/pull/249/commits/514d44406caf3165ee8beb46ce07b0e97ab01fd4)

src/data_source/sql.rs
jbearer: Fixed in [996c49b](https://github.com/EspressoSystems/hotshot-query-service/pull/249/commits/996c49bb55f47ff7f530bc3813239f1ee41fe11f)

src/data_source/sql.rs
jbearer: Added a check in [b919351](https://github.com/EspressoSystems/hotshot-query-service/pull/249/commits/b9193510bab679fe9ed3fa3cd98328855759ba04) that fails if these lines are commented out

src/data_source/sql.rs
```diff
@@ -608,134 +649,408 @@ impl VersionedDataSource for SqlDataSource {
...
     /// changes to the database is considered a fatal error, and this function may panic.
     async fn revert(&mut self) {
-        if !self.tx_in_progress {
-            // Nothing to do if there is no outstanding transaction.
-            return;
+        if self.tx_in_progress {
+            // If we're trying to roll back a transaction, something has already gone wrong and
+            // we're trying to recover. If we're unable to revert the changes and recover, all we
+            // can do is panic.
+            self.client
```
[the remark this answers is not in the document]
jbearer: Added in [880ef23](https://github.com/EspressoSystems/hotshot-query-service/pull/249/commits/880ef2371e7065e6e19857ea7bdd734bd041c342), actually caught some bugs in the file system backend

jbearer:
@sveitser I responded to all of your comments. Feel free to merge this Monday morning if you're satisfied with it, or push any fixes you want. Then you should be able to get started on the HotShot update

PR #251, opened 2023-11-13
Restructure modules so that it will make sense to have a second data source

### LINKED ISSUE #213: Make `QueryData` tests generic
These tests are almost completely black box. Make them complete generic of the type of the data source and use [`#[generic_tests]`](https://github.com/EspressoSystems/espresso-macros/blob/main/src/generic_tests.rs) to allow them to be instantiated with many different data sources.

jbearer:
* Rename `QueryData` to `FileSystemDataSource`, to contrast with
  `SqlDataSource` which we will be adding soon
* Move `FileSystemDataSource` to its own module under `data_source`,
  so we'll have `data_source::fs` and later `data_source::sql`
* Move some data source related types (LedgerLog and UpdateDataSource)
  under the `data_source` module
* Make data source tests generic and instantiate for `FileSystemDataSource`
* Restructure documentation, so crate-level documentation is now largely
  data source agnostic, and refers to the documentation of the individual
  data sources for data-source-specific things like extension and
  composability
    
Closes #213


sveitser:
I'm struggling a bit to understand current naming around "query data" when reading through the diff.

- There are `query_data` modules defining `XZYQueryData` structs. These types are used internal and externally for example in API responses. 
- There is a `query_data` method of `MockNetwork` and `MockNode` that returns `Arc<RwLock<_>>` of a `DataSource` impl.
- In the code often `qd` short for `query_data` (and sometimes just `query_data`) is used for the instances of types that implement `DataSource`.

Wonder if it would be a bit easier to read if as part of the rename from QueryData to DataSource we also renamed the usages in the code to `data_source` or something like that.

src/availability/data_source.rs
```diff
@@ -83,14 +83,14 @@ pub trait AvailabilityDataSource<Types: NodeType, I: NodeImplementation<Types>>
 where
     Block<Types>: QueryableBlock,
 {
-    type LeafStream: Stream<Item = LeafQueryData<Types, I>> + Send;
-    type BlockStream: Stream<Item = BlockQueryData<Types>> + Send;
+    type LeafStream: Stream<Item = LeafQueryData<Types, I>> + Unpin + Send;
```
sveitser, line 86: This is probably not so important but I  do wonder why we need unpin after the changes. Is it because some methods were made async? 

From what I can tell the trait bounds here are only used for tests (`cargo check` doesn't error if all of them are removed). For `LeafStream` the tests don't seem to require the trait bound either. So are we missing a test?

jbearer: That's weird. If I remove these trait bounds I get lots of errors in non-test code with `cargo check`, such as in the implementations of [`stream/leaves`](https://github.com/EspressoSystems/hotshot-query-service/pull/251/files/35c6ed9ab209f3f9f2c8aff5cad8c40a7cdab871#diff-445b1c6f33e6caf4e6b41096a67ef78c24caf4802c80acbb6e7d2c12973a2f12L167) and [`stream/blocks`](https://github.com/EspressoSystems/hotshot-query-service/pull/251/files/35c6ed9ab209f3f9f2c8aff5cad8c40a7cdab871#diff-445b1c6f33e6caf4e6b41096a67ef78c24caf4802c80acbb6e7d2c12973a2f12R223), where the returned types are required to be streams. Even if I only remove the `Send` and `Unpin` bounds, I get lots of errors. These traits are definitely required for the caller to be able to do anything useful with the returned value.

And yes, `Unpin` is needed because of things becoming async

src/data_source.rs
```diff
@@ -769,4 +266,70 @@ mod test {
...
+        // starting and ending bounds
+        do_range_test(&*qd, 1..=2, 1..3).await; // (inclusive, inclusive)
+        do_range_test(&*qd, 1..3, 1..3).await; // (inclusive, exclusive)
+        do_range_test(&*qd, 1.., 1..block_height).await; // (inclusive, unbounded)
+        do_range_test(&*qd, ..=2, 0..3).await; // (unbounded, inclusive)
+        do_range_test(&*qd, ..3, 0..3).await; // (unbounded, exclusive)
+        do_range_test(&*qd, .., 0..block_height).await; // (unbounded, unbounded)
+        do_range_test(&*qd, ExRange(0..=2), 1..3).await; // (exclusive, inclusive)
+        do_range_test(&*qd, ExRange(0..3), 1..3).await; // (exclusive, exclusive)
+        do_range_test(&*qd, ExRange(0..), 1..block_height).await; // (exclusive, unbounded)
```
nomaxg, line 298: Aren't these three tests equivalent to the first three after the ExRange wrapping?

nomaxg: In other words, what exactly are we testing here?

jbearer: The ranges being queried are the same but the ranges are being described in a different way. The first three are testing how we handle an inclusive lower bound and these three are testing how we handle an exclusive lower bound.

src/availability/data_source.rs
sveitser: Sorry I mean the `Unpin` trait bound, not the other trait bounds.

src/testing/mocks.rs
```diff
@@ -343,3 +347,19 @@ impl NodeImplementation<MockTypes> for MockNodeImpl {
...
+    + UpdateDataSource<MockTypes, MockNodeImpl>
+    + Send
+    + Sync
+    + Sized
+    + 'static
+{
+    type TmpData;
+
+    async fn create(node_id: usize) -> (Self, Self::TmpData);
+    async fn commit_version(&mut self);
```
nomaxg, line 364: May be out of scope - but would it be helpful for this to return a version number?

nomaxg (approved):
LGTM

src/testing/mocks.rs
jbearer: I'm not currently aware of a use case for this

src/availability/data_source.rs
jbearer: Ah, yes. There I only get errors in test code. However I think these errors demonstrate that having the `Unpin` bound is a useful thing for callers, and this is pretty normal for streams, so I would still be in favor of keeping the bound there

src/data_source/fs.rs
```diff
@@ -0,0 +1,673 @@
...
+}
+
+impl<Types: NodeType, I: NodeImplementation<Types>, UserData>
+    FileSystemDataSource<Types, I, UserData>
+where
+    Block<Types>: QueryableBlock,
+{
+    /// Create a new [FileSystemDataSource] with storage at `path`.
+    ///
+    /// If there is already data at `path`, it will be archived.
```
sveitser, line 201: This isn't part of this PR so we can maybe just create an issue for it if we want to fix it. 

I think it's better to fail if one calls `create` on a non-empty directory because it's likely a user error. It should be easy for the user of the code to make sure it's using an empty or non-existent directory. It's somewhat harder to figure out what files to move where if they were moved by mistake.

src/data_source/fs.rs
```diff
@@ -0,0 +1,673 @@
...
+mod test {
+    use super::super::data_source_tests;
+    use super::FileSystemDataSource;
+    use crate::testing::mocks::{MockNodeImpl, MockTypes};
+
+    // For some reason this is the only way to import the macro defined in another module of this
+    // crate.
+    use crate::*;
+
+    instantiate_data_source_tests!(FileSystemDataSource<MockTypes, MockNodeImpl, ()>);
```
sveitser, line 672: Silly question but where does `instantiate_data_source_tests` come from?

sveitser (approved):
LGTM

src/data_source/fs.rs
jbearer: It's defined by the [`#[generic_tests]`](https://github.com/EspressoSystems/hotshot-query-service/pull/251/files#diff-7136d6b9f26b655819ccf69416ccecb4d815aa64427a81900de400e450a44f15R32) macro

PR #265, opened 2023-11-20
Implement scaffolding for SQL data source

jbearer:
* Initialization and connection
* Migrations
* Documentation on extensibility and composability

Trying to break these reviews up into digestible chunks, so there is no "business logic" yet, just general SQL setup stuff. The implementation of the various queries will be the focus of the next PR.


migrations/V10__init_schema.sql
```diff
@@ -0,0 +1,47 @@
+CREATE TABLE header
```
nomaxg, line 1: Don't want to deviate you too far off course, but I'm wondering if about your thoughts on using an ORM? I personally find them a lot less daunting than modifying raw sql. In any case, this schema feels small enough to be manageable without an ORM.

migrations/V10__init_schema.sql
```diff
@@ -0,0 +1,47 @@
...
+    -- `serde_json`, and to run queries and create indexes on application-specific header fields
+    -- without having a specific column for those fields. In many cases, this will enable new
+    -- application-specific API endpoints to be implemented without altering the schema (beyond
+    -- possibly adding an index for performance reasons).
+    data JSONB NOT NULL
+);
+-- TODO this index should be unique, but HotShot does not currently give the application a way to
+-- put the block height or parent hash in the header, so headers are not guaranteed to be distinct.
+-- This will be fixed soon with the block header refactor.
+CREATE INDEX header_hash ON header (hash);
```
nomaxg, line 18: Can we still make this unique and add some rule to reconcile duplicates (i.e. store the header with the most recent timestamp)?

flake.nix
```diff
@@ -47,6 +47,8 @@
             cargo-udeps
             cargo-sort
             cmake
+
+            postgresql_15
```
sveitser, line 51: The default docker version is already 16 so wondering why we choose 15 here.

src/data_source/sql.rs
```diff
@@ -0,0 +1,987 @@
...
+/// the root of the crate.
+///
+/// ```
+/// # use hotshot_query_service::data_source::sql::{include_migrations, Migration};
+/// let migrations: Vec<Migration> = include_migrations!("$CARGO_MANIFEST_DIR/migrations").collect();
+/// assert_eq!(migrations[0].version(), 10);
+/// assert_eq!(migrations[0].name(), "init_schema");
+/// ```
+#[macro_export]
+macro_rules! include_migrations {
```
sveitser, line 76: In case you know, how is this different from https://docs.rs/refinery/0.8.11/refinery/macro.embed_migrations.html ?

src/data_source/sql.rs
```diff
@@ -0,0 +1,987 @@
...
+    // Check version uniqueness.
+    for (prev, next) in migrations.iter().zip(migrations.iter().skip(1)) {
+        if next <= prev {
+            return Err(Error::msg(format!(
+                "migration versions are not strictly increasing ({prev}->{next})"
+            )));
+        }
+    }
+
+    if default {
```
sveitser, line 124: I think this would be clearer with separate functions to check version spacing, check uniqueness, `validate_default_migrations` and `validate_custom_migrations` to remove the `default` boolean parameter that turns this function into one that does two things.

src/data_source/sql.rs
```diff
@@ -0,0 +1,987 @@
...
+
+    Ok(())
+}
+
+/// Merge two migration sequences.
+///
+/// Migrations in `right` replace migrations in `left` with the same version. Each of `left` and
+/// `right` is assumed to be the output of [`validate_migrations`]; that is, each is sorted by
+/// version and contains no duplicate versions.
+fn merge_migrations(
```
sveitser, line 143: This seems to be used only once to merge custom migrations into default migrations and is not exported. So I wonder if the code could be inlined instead or if we expect it to be useful elsewhere.

src/data_source/sql.rs
```diff
@@ -0,0 +1,987 @@
...
+        if let Some(user) = &config.user {
+            pgcfg.user(user);
+        }
+        if let Some(password) = &config.password {
+            pgcfg.password(password);
+        }
+        if let Some(database) = &config.database {
+            pgcfg.dbname(database);
+        }
+        let (mut client, connection) = pgcfg.connect_raw(tcp, NoTls).await?;
```
sveitser, line 482: Are we using `connect_raw` because the `connect` requires a tokio runtime?

src/data_source/sql.rs
```diff
@@ -0,0 +1,987 @@
...
+/// columns, indices) should usually be safe. Removing, altering, or renaming things should be done
+/// with extreme caution.
+///
+/// It is standard to store custom migrations as SQL files in a sub-directory of the crate. For ease
+/// of release and deploymenet, such directories can be embedded into a Rust binary and parsed into
+/// a list of [`Migration`] objects using the [`include_migrations`] macro.
+///
+/// It is also possible to take complete control over migrating the schema using
+/// [`Config::no_migrations`] to prevent the [`SqlDataSource`] from running its own migrations. The
+/// database adminstrator then becomes responsible for manually migrating the database, ensuring the
```
sveitser, line 325:
```suggestion
/// database administrator then becomes responsible for manually migrating the database, ensuring the
```

src/data_source/update.rs
```diff
@@ -126,3 +111,21 @@ where
...
+/// A data source with an atomic transaction-based synchronization interface.
+#[async_trait]
+pub trait VersionedDataSource {
+    type Error: Error + Debug;
+
+    /// Atomically commit to all outstanding modifications to the data.
+    ///
+    /// If this method fails, outstanding changes are left unmodified. The caller may opt to retry
+    /// or to erase outstanding changes with [`revert_version`](Self::revert_version).
+    async fn commit_version(&mut self) -> Result<(), Self::Error>;
```
sveitser, line 124: Since for postgres we can't revert back to older versions once we commit, should would it make more sense to call the trait functions `commit` and `revert` instead?

src/data_source/sql.rs
```diff
@@ -0,0 +1,987 @@
...
+    /// changes to the database is considered a fatal error, and this function may panic.
+    async fn revert_version(&mut self) {
+        if !self.tx_in_progress {
+            // Nothing to do if there is no outstanding transaction.
+            return;
+        }
+        // If we're trying to roll back a transaction, something has already gone wrong and we're
+        // trying to recover. If we're unable to revert the changes and recover, all we can do is
+        // panic.
+        self.client.batch_execute("ROLLBACK").await.unwrap();
```
sveitser, line 606: Something like `expect("DB rollback succeeds.")` might help the person looking at the panic.

sveitser:
Could you elaborate a bit on what prompted the choice of `tokio-postgres`? Looking at this diff it seems we have custom code arising from the way we want to do migrations, using async-std as well as the way we want to handle database transactions and rollbacks? 

Since there are other libraries and for example https://github.com/launchbadge/sqlx claims to work with async-std it would be good to have the rationale for the choice of library elaborated somewhere.

migrations/V10__init_schema.sql
jbearer: Not a bad idea. I don't think we get the full leverage of using an ORM since we aren't really doing complicated things with the query results. We're mainly just loading a bunch of data and passing it straight through to the client. But it would handle a lot of the boilerplate parsing and query generation. I'll make an issue and we can do this as a follow-up without changing the interface.

jbearer: https://github.com/EspressoSystems/hotshot-query-service/issues/270

migrations/V10__init_schema.sql
jbearer: It's annoying because then you have to check if a header with the same hash already exists and use `UPDATE` instead of `INSERT` (`INSERT` will fail if there is a unique index with an already existing entry). I think it's simpler to leave it the way it is and just add the unique constraint once we have block height in the header

src/data_source/sql.rs
jbearer: `embed_migrations` gives you a [`Runner`](https://docs.rs/refinery/latest/refinery/struct.Runner.html#), but I just wanted a `Vec<Migration>` so I could add other migrations and modify the order and whatnot (see `validate_migrations`). We will use this capability in the sequencer to insert additional migrations to create indices for sequencer-specific queries.

src/data_source/sql.rs
jbearer: I don't know if it will be useful anywhere, but I think it still makes sense as a separate function. It's a self-contained bit of logic and if I inlined it into `connect`, the nesting and complexity of that function would get a bit out of hand IMO. However, I can make the comment and names more explicit (e.g. `default` and `custom` instead of `left` and `right`)

src/data_source/sql.rs
```diff
@@ -0,0 +1,987 @@
...
+    }
+}
+
+impl UpdateStatusData for SqlDataSource {
+    fn metrics(&self) -> Box<dyn Metrics> {
+        todo!()
+    }
+}
+
+pub struct Transaction<'a> {
```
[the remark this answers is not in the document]
jbearer: lol. An earlier iteration of this required `Transaction` to be `'static`, so I implemented this wrapper around client. Then apparently I got rid of that restriction and added the lifetime back in, but now the whole reason for having a custom `Transaction` type is gone...I think I can switch it back to `tokio_postgres::Transaction` now

src/data_source/update.rs
jbearer: Perhaps. `atomic_store` doesn't let you revert to old versions either, so the behavior isn't really any different. But I think your names are better/less confusing

src/data_source/sql.rs
```diff
@@ -0,0 +1,987 @@
...
+///         }
+///     });
+///
+///     Ok(app)
+/// }
+/// ```
+#[derive(Debug)]
+pub struct SqlDataSource {
+    client: Client,
+    tx_in_progress: bool,
```
nomaxg, line 461: Why not Option<Transaction<'>> instead? That way we can simply drop the txn in revert_version and we wouldn't need an extra variable to track whether we have a txn in progress.

jbearer:
I chose `tokio-postgres` because it was the only good Postgres client I could find. There is a [synchronous client](https://docs.rs/postgres/0.19.7/postgres/), but we want to be async. There is [async postgres](https://docs.rs/async-postgres/latest/async_postgres/), which is runtime-agnostic but is really just a wrapper around `tokio-postgres` (which is also pretty much runtime-agnostic). I've used `async-postgres` before and it pins old versions of some dependencies which makes it a bit annoying to work with.

I hadn't seen sqlx. It looks good. `refinery` doesn't support it, but it seems to have its own built-in migrations support, so it would probably be an improvement. OTOH once I remove the custom `Transaction` type as you suggested above I think a lot of the `tokio-postgres` rough edges will be gone.

@sveitser WDYT about merging this PR as-is (after addressing other comments) and switching to sqlx later on, vs stopping and switching to sqlx now? I'm a bit worried about getting this project merged and integrated into the sequencer before things get too divergent (especially since there is a HotShot update kind of wrapped up in this change), and I think we can come back and switch to sqlx later without breaking any interfaces. But I am also sympathetic to concerns about taking on more tech debt.

src/data_source/sql.rs
jbearer: That's what I wanted to do originally, but because of the lifetime in `Transaction<'a>` you can't store a transaction here in the struct itself, as that would be self-referential.

src/data_source/sql.rs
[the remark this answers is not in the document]
jbearer: Oh no, I remember why this was necessary. https://github.com/EspressoSystems/hotshot-query-service/pull/265#discussion_r1406455649

`tokio_postgres::Transaction` executes a `BEGIN` command every time it is created and a `COMMIT` every time it is dropped. But to avoid a self-reference I create and drop these `Transaction` wrappers all the time, and only create/commit a transaction in the database when `commit_version`/`revert_version` is called. So I think this wrapper is still necessary sadly

nomaxg (approved):
LGTM

flake.nix
jbearer: 15 is the newest version in nixpkgs. If you just do `postgresql` you get an even older version (14.9)

sveitser: There should be a _16, probably need to `nix flake update` if it's not there.

https://search.nixos.org/packages?channel=unstable&show=postgresql_16&from=0&size=50&sort=relevance&type=packages&query=postgresql_16

sveitser (approved):
@jbearer thanks for the elaborations. Totally fine with merging it. I think it shouldn't be too difficult to change the library in case we want that later.

flake.nix
jbearer: Ah that did the trick

src/data_source/sql.rs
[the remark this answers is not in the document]
jbearer: Added a comment explaining this in [a1423b1](https://github.com/EspressoSystems/hotshot-query-service/pull/265/commits/a1423b142a29f313cbe9a4310725d9f5972333e8)

PR #268, opened 2023-11-21
fix: QueryableBlock return an owned Transaction

### LINKED ISSUE #266: `QueryableBlock::transaction_with_proof` is lending a `Transaction` reference
Return type includes `&Self::Transaction`:

https://github.com/EspressoSystems/hotshot-query-service/blob/159eb6c5a0c4775089ff1954d4641d72b18df5ec/src/availability/query_data.rs#L102-L105

This is a problem because I have only the raw tx bytes, not a proper `Transaction` struct. I could construct on one the fly but then I would be returning a reference to a temporary value. Basically, this API forces the implementation to store all transactions as `Transaction`.

Not sure what the right solution is here. I see why you want to return a reference---we should not be cloning a (possibly large) transaction. Do we need to modify `Transaction` itself? Maybe instead of owned `Vec<u8>` it could have a `&[u8]` with lifetime.

ggutoski:
close #266

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,18 +24,19 @@ additional, application-specific features (such as a JSON-RPC service in the cas
 application).
 
 ## Features
-* Provides a query interface to a persistent store containing the history of a blockchain
+
+- Provides a query interface to a persistent store containing the history of a blockchain
   (represented as a chain of HotShot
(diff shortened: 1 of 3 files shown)


jbearer (approved):
Looks good! Thanks for the README improvements!

PR #303, opened 2023-12-11
fix: update of outdated link

bulikos:
During the documentation browsing, I discovered a link pointing to non-existing resource. This PR fixes the issue.


jbearer (approved):
Thanks @bulikos !

PR #307, opened 2023-12-12
Add a feature to reset the SQL database; streamline and extend the public query API

jbearer:
This probably should have been two separate PRs, sorry about that. Got a little carried away. If you look at the diff for each commit separately, hopefully it is pretty straightforward.

### DIFF
--- a/src/data_source/sql.rs
+++ b/src/data_source/sql.rs
@@ -48,7 +48,7 @@ use std::{
 };
 use tokio_postgres::{
     types::{BorrowToSql, ToSql},
-    Client, NoTls, Row, RowStream, ToStatement,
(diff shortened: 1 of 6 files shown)


src/data_source/sql.rs
```diff
@@ -301,6 +334,15 @@ impl Config {
 /// # ;
 /// ```
 ///
+/// ## Resetting
+///
+/// In general, resetting the database when necessary is left up to the adminstrator. However, for
```
sveitser, line 339:
```suggestion
/// In general, resetting the database when necessary is left up to the administrator. However, for
```

sveitser (approved):
LGTM

PR #319, opened 2023-12-20
Feat/da fetching take 4

### LINKED ISSUE #217: Spawn a task to recover blocks from HotShot
We should use a (multi-)producer (single-)consumer model, so that anywhere in the query service can enqueue a request to recover a block, but only one task is responsible for handling the requests. This task can thus deduplicate requests for the same block coming from different producers. The consumer task can delegate the actual handling of the requests to a pool of worker tasks if necessary, although this will be primarily I/O-bound, so this may not be necessary.

The consumer task will use [the HotShot API](https://github.com/EspressoSystems/HotShot/issues/1640) to do the actual fetching.

There should be an API for launching this task, which the application is expected to call at start-up. The query data and API modules will then need to be initialized with handles to the task (i.e. channel endpoints they can use to send requests and receive results).

Anywhere we discover a block is missing (ie when loading query data at startup, when receiving a leaf from HotShot with no corresponding block, or when looking up a block in response to a query) we should enqueue a request to fetch that block. The query data should subscribe to a stream of fetched blocks and update itself accordingly.

jbearer:
# This PR does:

Reconfigure the architecture of the availability data source as suggested by @nyospe. Specifically:
* Instead of failing, requests for objects return a future `Fetch` that will either resolve immediately or block until the object can be retrieved from an external DA provider
* Move the existing "data source" implementations to a new, lower abstraction level: the storage layer
* Create a new layer which wraps a storage implementation and a remote DA provider. It fetches from the provider whenever requests to the storage layer fail
* Closes #217 

While the diff is very large, almost everything in it is a reaction to that fairly simple conceptual change:
* `FileSystemDataSource` and `SqlDataSource` got moved almost wholesale to new modules under `data_source/storage`
* A new API module was created to host the proposal endpoints, which didn't fit in either the availability module (because they are not immutable) or the status module (because they require persistence)
* A fair amount of new code was added in `fetching.rs` and `notifier.rs` to implement (passive, for now) asynchronous fetching

# This PR does not:
* Implement fetching from a provider. This will come next, once we're satisfied with the architecture, since a fetching implementation can just be dropped into the hole in this PR: `Fetchable::active_fetch`
* Spawn a background task to look for missing objects and fetch them. This can also be added on after the fact without affecting the architecture. It will eventually be needed to make catchup work, since otherwise we rely on client requests to trigger fetches of missing data.
* An optimization I would like to add is keeping the block height up-to-date in memory. This will let us bypass the storage layer completely when requesting a block or leaf which hasn't been produced yet, and instead just wait for a notification, like we did in the old implementation. This is especially common for streams, which after a while are always fetching the next object to be produced.


src/data_source/notifier.rs
```diff
@@ -0,0 +1,363 @@
...
+use async_std::sync::Arc;
+use derivative::Derivative;
+use futures::future::{BoxFuture, FutureExt};
+use std::{
+    future::IntoFuture,
+    sync::atomic::{AtomicBool, Ordering},
+};
+
+pub trait Predicate<T>: 'static + Send + Sync + Fn(&T) -> bool {}
+impl<F, T> Predicate<T> for F where F: 'static + Send + Sync + Fn(&T) -> bool {}
```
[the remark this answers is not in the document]
nomaxg: Can you explain this Fn(&T) bound?

src/data_source/notifier.rs
```diff
@@ -0,0 +1,363 @@
...
+    closed: Arc<AtomicBool>,
+}
+
+impl<T> Subscriber<T> {
+    fn is_closed(&self) -> bool {
+        // A subscriber can be closed because it was explicitly closed by its receiver (e.g. the
+        // receiver was dropped)
+        self.closed.load(Ordering::Relaxed) ||
+        // Or because it has already been notified, which `take`s the oneshot sender.
+        self.sender.is_none()
```
nomaxg, line 103: Small nit but maybe a bit more efficient to put the is_none() check first?

src/data_source/fetching.rs
```diff
@@ -0,0 +1,885 @@
...
+                        // If `passive_fetch` returns `None`, it means the notifier was dropped
+                        // without ever sending a notification. In this case, the correct behavior
+                        // is actually to block forever (unless the `Fetch` itself is dropped),
+                        // since the semantics of `Fetch` are to never fail. This is analogous to
+                        // fetching an object which doesn't actually exist: the `Fetch` will never
+                        // return.
+                        tracing::warn!(
+                            "notifier dropped without satisfying request {req:?}, fetch will block"
+                        );
+                        future::pending().await
```
nomaxg, line 503: Hmm this feels like strange behavior - I understand the notion of blocking forever to be consistent with the semantics of fetch, but is there ever a practical reason to block forever in this case?

src/data_source/notifier.rs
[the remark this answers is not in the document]
jbearer: This trait is just an alias for `Fn(&T) -> bool` (with a few extra things for concurrency)

src/data_source/notifier.rs
jbearer: Fair enough

src/data_source/fetching.rs
jbearer: Yeah you're right, panic is probably better here

nomaxg (approved):
LGTM

PR #327, opened 2023-12-22
Upgrade to stable HotShot

jbearer:


jbearer:
I think the point of the 0.5.3 release is to be something stable for Gibraltar, so that the consensus team can start merging more breaking changes into main. And if we need to patch Gibraltar we will branch off of 0.5.3 and do another tag. I could have misunderstood. @shenkeyao or @elliedavidson could you clarify?

elliedavidson:
> I think the point of the 0.5.3 release is to be something stable for Gibraltar, so that the consensus team can start merging more breaking changes into main. And if we need to patch Gibraltar we will branch off of 0.5.3 and do another tag. I could have misunderstood. @shenkeyao or @elliedavidson could you clarify?

Yes, this is correct.  We have other changes coming in that might break things, and we want to keep those out of the Gibraltar release.

elliedavidson (approved):

PR #344, opened 2024-01-10
Add payload metadata parameter to QueryablePayload methods

### LINKED ISSUE #328: `QueryableBlock` transaction look-up functions should take a new header arg
Distilled from [zulip](https://espresso.zulipchat.com/#narrow/stream/330967-Sequencer/topic/.E2.9C.94.20block.20payload.20API.20for.20metadata).

The new functions will look something like:
```rust
   fn transaction_with_proof(
        &self,
        index: &Self::TransactionIndex,
        metadata: &Self::Metadata
    ) -> Option<(Self::Transaction, Self::InclusionProof)>;
```
Or maybe you want the new arg to `impl BlockHeader` or something.

jbearer:
This required a HotShot update, which in turn resulted in a number of minor changes related to proposer IDs/signature keys.

Closes #328

### DIFF
--- a/src/availability.rs
+++ b/src/availability.rs
@@ -10,16 +10,17 @@
 // You should have received a copy of the GNU General Public License along with this program. If not,
 // see <https://www.gnu.org/licenses/>.
 
-use crate::{api::load_api, Payload, QueryError, QueryResult};
+use crate::{api::load_api, Payload, QueryError, QueryResult, SignatureKey};
 use clap::Args;
 use derive_more::From;
 use futures::{FutureExt, Stream, StreamExt, TryFutureExt, TryStreamExt};
+use hotshot::types::SignatureKey as _;
 use hotshot_types::traits::{node_implementation::NodeType, signature_key::EncodedPublicKey};
 use serde::{Deserialize, Serialize};
 use snafu::{OptionExt, ResultExt, Snafu};
 use std::fmt::Display;
 use std::path::PathBuf;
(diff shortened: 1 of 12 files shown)


src/availability.rs
```diff
@@ -288,28 +290,43 @@ where
...
 }
 
+fn proposer_param<Types: NodeType>(
+    req: &RequestParams,
+    param: &str,
+) -> Result<SignatureKey<Types>, Error> {
+    // The HotShot signature key trait temporarily lacks the trait bounds required to convert
+    // directly from TaggedBase64. As a workaround, we parse the TaggedBase64 as an
+    // EncodedPublicKey and then decode to the actual signature key type.
+    let encoded: EncodedPublicKey = req.blob_param(param)?;
```
nomaxg, line 326: can we attach an issue to this?

src/availability/query_data.rs
```diff
@@ -149,17 +163,19 @@ pub trait QueryablePayload: traits::BlockPayload {
     /// Get the transaction with a given hash, if it is in the block.
     fn transaction_by_hash(
         &self,
+        meta: &Self::Metadata,
```
nomaxg, line 166: I'm curious, what exactly is this metadata?

jbearer: It's extra information produced when a block is built, which doesn't go in the payload itself but is required to interpret the payload: https://hotshot.docs.espressosys.com/hotshot/traits/trait.BlockPayload.html#tymethod.from_transactions.

For the sequencer, this is the namespace table, which ultimately goes in the block header.

PR #351, opened 2024-01-12
Make Postgres storage more robust

jbearer:
* Detect silent tokio-postgres errors (🙄) by looking at the number of rows modified
* Retry when a statement fails to execute or have an effect


nomaxg (approved):
LGTM

PR #365, opened 2024-01-16
Implement fetching from a peer query service

### LINKED ISSUE #296: Payloads
* add payload routes corresponding to each block/leaf route

### LINKED ISSUE #326: Implement DA fetching
#319  address #217 , but leaves the actual implementation of fetching unimplemented, so for now the spawned task blocks forever. We should define a trait so that we can have various different sources of data, and implement this trait for now with a client that queries another instance of the query service. Later on we can also implement it by querying the HotShot CDN or p2p network directly.

Also need to add tests for catchup and missing data, once fetching is supported.

### LINKED ISSUE #337: Implement DA fetching from a provider
Implement fetching from a provider. This will come next, once we're satisfied with the architecture, since a fetching implementation can just be dropped into the hole in this PR: Fetchable::active_fetch

jbearer:
This PR implements `QueryServiceProvider` to fetch data from a peer when missing, as well as number of tests for fetching. There are also numerous changes to the fetching infrastructure in `data_source/fetching.rs` and the SQL storage backend, which were necessitated as I ran into various places where the previous version just didn't work. Nothing about the fundamental design is changed, though, just a lot of details.

I also ended up adding new endpoints for fetching payloads on their own (#296) which was a feature we wanted anyways and turned out to be important for implementing block fetching in a reasonable way (if we already have the header we don't want to refetch it).

Design documentation in [doc/async-retrieval.md](https://github.com/EspressoSystems/hotshot-query-service/blob/feat/fetching-provider/doc/async-retrieval.md).

I have annotated in comments on this PR the most important and/or confusing changes. Will schedule some time next week to go through it in person.

Closes #296 
Closes #326 
Closes #337


examples/simple-server.rs
```diff
@@ -28,15 +28,15 @@ use hotshot::{
     HotShotInitializer, Memberships, Networks, SystemContext,
```
jbearer, line 28: Changes in here are just from interface changes in HotShot and the data sources

migrations/V10__init_schema.sql
```diff
@@ -49,5 +49,6 @@ CREATE TABLE transaction
     -- make use of the transaction index in its own SQL queries.
     index JSONB NOT NULL
 );
+ALTER TABLE transaction ADD CONSTRAINT transaction_pk PRIMARY KEY (block_height, index);
```
jbearer, line 52: Changed this to have an explicit primary key to make upserts more intuitive, and prevent duplicate transactions from the same block being inserted more than once

src/availability/data_source.rs
```diff
@@ -35,27 +35,56 @@ use std::{cmp::Ordering, error::Error, fmt::Debug, ops::RangeBounds};
...
+#[derivative(Ord = "feature_allow_slow_enum")]
+#[derivative(
+    Copy(bound = ""),
+    Debug(bound = ""),
+    PartialEq(bound = ""),
+    Eq(bound = ""),
+    Ord(bound = ""),
+    Hash(bound = "")
+)]
+pub enum BlockId<Types: NodeType> {
```
jbearer, line 67: We have a separate type for `BlockId` now, rather than using `ResourceId`, because there are more ways to query for a block than for a leaf

src/availability/query_data.rs
```diff
@@ -257,47 +265,18 @@ pub struct BlockQueryData<Types: NodeType> {
...
-            qc.data.leaf_commit == leaf.commit(),
-            InconsistentLeafSnafu {
-                leaf: leaf.commit(),
-                qc_leaf: qc.data.leaf_commit
-            },
-        );
-        Ok(Self {
-            hash: leaf.block_header.commit(),
-            header: leaf.block_header.clone(),
+    pub fn new(header: Header<Types>, payload: Payload<Types>) -> Self {
```
jbearer, line 268: Simplified constructor, starting to think of a block more as just a combination of header and payload, rather than its own thing

src/data_source/fetching.rs
```diff
@@ -340,22 +365,66 @@ where
     Types: NodeType,
 {
     storage: S,
+    height: u64,
```
jbearer, line 368: Did the thing I mentioned before where we keep the height in memory, this is actually pretty important for efficiency particularly for subscription streams, where we do a simple and elegant but not inherently performant thing: generate the subscription from basically an infinite stream of `Fetch` requests. This could be really bad because it would generate a DB query for every single entry in the stream, even though we know that once we reach the end of the current chain, subsequent entries will not actually be in the database and will instead be received by the event stream. This in-memory height allows these fetches to short circuit before they hit the DB, making this simple design just about as efficient as a more complicated thing with separate logic for blocks before and after the end of the chain.

src/data_source/fetching.rs
```diff
@@ -563,19 +660,21 @@ where
...
-    /// leaves that are ever produced.
+    /// If we do trigger an active fetch for an object, the provided callback will be called if and
+    /// when the fetch completes successfully. The callback should be responsible for notifying any
+    /// passive listeners that the object has been retrieved. If we do not trigger an active fetch
+    /// for an object, this function does nothing.
+    ///
+    /// In either case, as long as the requested object does in fact exist, we will eventually
+    /// receive it passively, since we will eventually receive all blocks and leaves that are ever
+    /// produced. Active fetching merely helps us receive certain objects sooner.
     async fn active_fetch<S, P>(
```
jbearer, line 671: Biggest change in all of this is how active fetching works. Basically because active fetching happens in a separate task, and may not happen at all depending on a decision taken after the task is already spawned, it was a real pain to try to have active fetch return the fetched object. It's not really necessary either, since we eventually get the object via the `Notifier`. It ended up being a lot simpler, IMO, to just use callbacks as needed to post-process fetched objects

src/data_source/fetching.rs
```diff
@@ -641,22 +738,84 @@ where
...
+                // created.
+                tracing::info!("not fetching leaf {n} because height is only {height}");
+                return;
+            }
+
+            let fetcher = fetcher.clone();
+            fetcher.leaf_fetcher.clone().spawn_fetch(
+                n.into(),
+                fetcher.provider.clone(),
+                once(LeafCallback::Leaf { fetcher }).chain(callbacks),
```
jbearer, line 793: We use a dedicated type/trait for callbacks now, rather than a closure, because callbacks actually have an identity. More about this in `src/fetching.rs`, but basically this is to ensure that all callbacks run, for example in the case where we're fetching a leaf and the corresponding block at the same time, even though we only actually do the fetch once.

src/data_source/storage/sql.rs
```diff
@@ -515,19 +546,39 @@ where
...
+        // ORDER BY ASC ensures that if there are duplicate transactions, we return the first
         // one.
         let query = format!(
             "SELECT {BLOCK_COLUMNS}, t.index AS tx_index
                 FROM header AS h
                 JOIN payload AS p ON h.height = p.height
                 JOIN transaction AS t ON t.block_height = h.height
                 WHERE t.hash = $1
-                ORDER BY t.id ASC
+                ORDER BY (t.block_height, t.index) ASC
```
jbearer, line 581: Now we see the importance of this schema change: this deterministic primary key gives us the first transaction in chronological order, whereas the old serial column would just order by when this particular DB first saw each transaction (possibly out of order)

src/data_source/storage/sql.rs
```diff
@@ -565,24 +616,24 @@ where
             serde_json::to_value(&leaf.leaf().block_header).map_err(|err| QueryError::Error {
                 message: format!("failed to serialize header: {err}"),
             })?;
-        tx.execute_one_with_retries(
-            "INSERT INTO header (height, hash, data) VALUES ($1, $2, $3)",
-            [
+        tx.upsert(
```
jbearer, line 619: Changed all inserts/updates to atomic upserts, I just felt with all the new concurrency it was too error prone to try to reason about when it was possible for the same resource to be inserted twice or not.

src/fetching/provider.rs
```diff
@@ -0,0 +1,75 @@
...
+//! availability service, including the HotShot CDN, the data availability committee which is
+//! responsible for providing a particular payload, or another instance of this same query service.
+//!
+//! This module defines an abstract interface [`Provider`], which allows data to be fetched from any
+//! data availability provider, as well as various implementations for different data sources,
+//! including:
+//! * [`QueryServiceProvider`]
+//!
+//! We also provide combinators for modularly adding functionality to existing fetchers:
+//! * [`AnyProvider`]
```
jbearer, line 34: This is copied from another branch and doesn't actually exist yet, but I intend to add it soon. Basically a provider that holds a list of other providers and tries each of them sequentially until a fetch succeeds. This will allow us to configure the sequencer with a list of peer query services to fetch from, and eventually even combine multiple different kinds of data sources, like peers, CDN, DA committee, and VID

doc/async-retrieval.md
```diff
@@ -0,0 +1,175 @@
...
+* Requests that should block until data becomes available
+* Proactive retrieval of missing data to aid future requests
+
+The first two are enabled by the `Fetch` abstraction. `Fetch` is a special kind of future
+representing either data that has been retrieved already, or a pending attempt to fetch data. When
+awaited, `Fetch` will return immediately when the data is available, or block until it becomes
+available. But it is also possible to check if the data is available now and fail if not, using
+`Fetch::try_resolve()`.
+
+The two use cases of `Fetch` can be characterized as "on demand" fetching: we only look for data
```
tbro, line 50: Seems confusing as there are actually three cases, the third of which is explained at the end of the paragraph. "*These* two use cases..." seems clearer to me.

tbro:
LGTM. I don't have an approval button though :cry:

src/availability.rs
```diff
@@ -209,6 +215,33 @@ where
...
+            async move {
+                let id = if let Some(height) = req.opt_integer_param("height")? {
+                    BlockId::Number(height)
+                } else if let Some(hash) = req.opt_blob_param("hash")? {
+                    BlockId::PayloadHash(hash)
+                } else {
+                    BlockId::Hash(req.blob_param("block-hash")?)
+                };
+                state.get_payload(id).await.context(FetchBlockSnafu {
+                    resource: id.to_string(),
```
nomaxg, line 228: Does it makes sense to include some of the `req` context in the error? I don't recall whether tide disco handles this

src/data_source/fetching.rs
```diff
@@ -340,22 +365,66 @@ where
...
 impl<Types, S, P> Fetcher<Types, S, P>
 where
     Types: NodeType,
+    S: NodeDataSource<Types>,
 {
-    fn new(storage: S, provider: P) -> Self {
+    async fn new(storage: S, provider: P) -> Self {
+        // Get the height from storage if possible. If not, it's fine: we'll update this as soon as
+        // we see a new block or leaf anyways.
+        let height = storage.block_height().await.unwrap_or_else(|err| {
```
nomaxg, line 413: Could this fail in a recoverable way? Just wondering if a panic is more appropriate if failing here means that storage is botched

doc/async-retrieval.md
jbearer: Ah, it was referring to the fact the only the first two of the 3 listed use cases are related to `Fetch`. I see how this is confusing, let me reword

src/availability.rs
jbearer: We include the ID of the resource we failed to fetch (which is derived from `req`). Anything else you'd want to see?

src/data_source/fetching.rs
jbearer: Do nothing and returning 0 _is_ recoverable: it is never _incorrect_ to have a height less than the true height -- this can happen for many reasons including our attached HotShot node lagging, and we will update height as soon as our HotShot client updates.

As for whether storage is botched, it could be, but it could also be a transient failure that will resolve itself, like we temporarily lost connectivity to the DB server.

nomaxg: gotcha makes sense

src/availability.rs
nomaxg: hmmm nothing in particular comes to mind!

src/fetching.rs
```diff
@@ -0,0 +1,135 @@
...
+    /// check that the resource is being fetched requires an exclusive lock, and we do not want to
+    /// block the caller, which might be on the critical path of request handling.
+    ///
+    /// Note that while callbacks are allowed to be async, they are executed sequentially while an
+    /// exclusive lock is held, and thus they should not take too long to run or block indefinitely.
+    pub fn spawn_fetch<Types>(
+        &self,
+        req: T,
+        provider: impl Provider<Types, T> + 'static,
+        callbacks: impl IntoIterator<Item = C> + Send + 'static,
```
nomaxg, line 81: Does it make sense to somehow limit the number of callbacks?

nomaxg (approved):
LGTM - one last question

src/fetching.rs
jbearer: I don't think it's necessary. Callbacks are deduplicated, plus the list of callbacks is determined entirely by our own internal logic, it's not subject to user input

PR #379, opened 2024-01-24
Follow-up for async fetching, changes needed for sequencer integration

jbearer:
See commit messages for details

### DIFF
--- a/src/fetching.rs
+++ b/src/fetching.rs
@@ -23,10 +23,13 @@
 
 use async_std::{
     sync::{Arc, Mutex},
-    task::spawn,
+    task::{sleep, spawn},
 };
 use derivative::Derivative;
-use std::collections::{hash_map::Entry, BTreeSet, HashMap};
+use std::{
+    collections::{hash_map::Entry, BTreeSet, HashMap},
+    time::Duration,
+};
 
 pub mod provider;
 pub mod request;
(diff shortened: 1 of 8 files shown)


src/fetching.rs
```diff
@@ -109,7 +138,25 @@ impl<T, C> Fetcher<T, C> {
...
+                //
+                // To understand why it is ok to retry indefinitely, think about manual
+                // intervention: if we don't retry, or retry with a limit, we may require manual
+                // intervention whenever a query service fails to fetch a resource that should exist
+                // and stops retrying, since it now may never receive that resource. With indefinite
+                // fetching, we require manual intervention only when active fetches are
+                // accumulating because a peer which _should_ have the resource isn't providing it.
+                // In this case, we would require manual intervention on the peer anyways.
+                tracing::warn!("failed to fetch {req:?}, will retry in {retry_delay:?}");
+                sleep(retry_delay).await;
```
nomaxg, line 158: In the future it might be nice to have a backoff setting

src/fetching/provider/any.rs
```diff
@@ -0,0 +1,254 @@
...
+        P: Provider<Types, PayloadRequest> + Provider<Types, LeafRequest> + Debug + 'static,
+    {
+        let provider = Arc::new(provider);
+        self.payload_providers.push(provider.clone());
+        self.leaf_providers.push(provider);
+        self
+    }
+
+    /// Add a sub-provider which fetches blocks.
+    pub fn with_block_provider<P>(mut self, provider: P) -> Self
```
nomaxg, line 128: I'm curious, how would a generic version of this look, I am spacey right now with Rust generics (thanks Go lol), but would it be possible to have some function like `with_provider<P, T>` that adds to a generic providers map like:
```
providers: Vec<HashMap<TypeKey<T>, P> 
```

Anyways, this isn't important because we only have two sub-providers but I was curious.

src/fetching.rs
jbearer: Good idea, I implemented a simple exponential backoff bounded by a maximum (which is configurable), this lets us try a few times very quickly but then back off to a very slow retry rate to reduce spam

nomaxg: nice!

src/fetching/provider/any.rs
jbearer: Yeah, I think you could do something like this with [`TypeId`]. But it would be messy, since you don't know the types of `T` and `P` at the data structure level (only in this particular method) you would have to do something like `HashMap<TypeId, Arc<dyn Provider>>`, and then when you want to use the provider you would have to downcast it to a specific request type.

PR #390, opened 2024-02-02
Chaos testing: Implement a query storage that relies purely on peer data

### LINKED ISSUE #338: Spawn a background task to look for missing DA objects
Spawn a background task to look for missing objects and fetch them. This can also be added on after the fact without affecting the architecture. It will eventually be needed to make catchup work, since otherwise we rely on client requests to trigger fetches of missing data.

### LINKED ISSUE #378: Chaos testing: Implement a query storage that relies purely on peer data
Since local query storage is a trait, impl that without real storage. Run instance of query service that doesn’t have persistence at all, but gets data for peer. This tests how the system responds when data is missing.

### LINKED ISSUE #386: Change `<FetchingDataSource as {NodeDataSource,StatusDataSource}>::block_height` to use in-memory height
We store the block height in memory, but these queries are still reading from the database. Using the in memory height should decrease latency and take a decent amount of pressure off the database, since these are very common queries

jbearer:
Implement `NoStorage` which implements the availability storage interface by returning errors for every query. This lets us really stress test fetching. With a few modifications to the tests and new features we had already planned on implementing, we are able to run the availability tests and have them pass without any persistent storage!

**Test modifications**
* We don't enforce that a query for a non-unique payload or transaction
  returns the _first_ matching object, if the object needs to be fetched.
  We do guarantee this for objects which are stored locally, but there's
  no good way to do it when fetching the object, since the proactive fetcher
  may provide a matching object in any order. This is not a property anybody relies on anyways.
* Moved the proposer queries to a separate test module. Since they are no
  longer part of the availability API, these queries do not support fetching
  and thus don't work with [`NoStorage`]

**New features**
* Use in-memory height when responding to `block-height` queries. This makes the queries faster and puts less pressure on the database. It was needed here because block height would simply not be accurate for the `NoStorage` tests if we tried fetching it from the (fake) database.
* Background task to identify and fetch missing objects. This always decreases the probability that a query will trigger an active fetch or fail due to data being missing, which is very nice. It is particularly important for `NoStorage`, because it allows us to succeed even in queries that use passive fetches.

One more thing I ~want to do~did before making this ready for review: add a timeout for availability endpoint handlers: if the requested resource is not available immediately, wait a short time on par with a normal HTTP response (maybe 500ms or so) to see if the object arrives on time. This will allow us to succeed in cases where an object was missing but we triggered an active fetch for it which quickly succeeds, which should reduce the error rate a lot for nodes with missing or unreliable storage. This in turn may allow us to run the proactive fetcher less frequently, which further reduces load on the CPU and database.

Closes #378 
Closes #386 
Closes #338

### DIFF
--- a/src/availability.rs
+++ b/src/availability.rs
@@ -28,13 +28,13 @@
 
 use crate::{api::load_api, Payload};
 use clap::Args;
+use cld::ClDuration;
 use derive_more::From;
 use futures::{FutureExt, StreamExt, TryFutureExt};
 use hotshot_types::traits::node_implementation::NodeType;
 use serde::{Deserialize, Serialize};
 use snafu::{OptionExt, Snafu};
-use std::fmt::Display;
-use std::path::PathBuf;
+use std::{fmt::Display, path::PathBuf, str::FromStr, time::Duration};
 use tide_disco::{api::ApiError, method::ReadState, Api, RequestError, StatusCode};
 
 pub(crate) mod data_source;
@@ -49,6 +49,19 @@ pub struct Options {
     #[arg(long = "availability-api-path", env = "HOTSHOT_AVAILABILITY_API_PATH")]
     pub api_path: Option<PathBuf>,
 
+    /// Timeout for failing requests due to missing data.
+    ///
+    /// If data needed to respond to a request is missing, it can (in some cases) be fetched from an
+    /// external provider. This parameter controls how long the request handler will wait for
+    /// missing data to be fetched before giving up and failing the request.
+    #[arg(
+        long = "availability-fetch-timeout",
+        env = "HOTSHOT_AVAILABILITY_FETCH_TIMEOUT",
+        default_value = "500ms",
+        value_parser = parse_duration,
+    )]
+    pub fetch_timeout: Duration,
+
     /// Additional API specification files to merge with `availability-api-path`.
     ///
     /// These optional files may contain route definitions for application-specific routes that have
@@ -59,6 +72,19 @@ pub struct Options {
         value_delimiter = ','
     )]
     pub extensions: Vec<toml::Value>,
+}
+
+#[derive(Clone, Debug, Snafu)]
+struct ParseDurationError {
+    reason: String,
+}
+
+fn parse_duration(s: &str) -> Result<Duration, ParseDurationError> {
+    ClDuration::from_str(s)
+        .map(Duration::from)
+        .map_err(|err| ParseDurationError {
+            reason: err.to_string(),
+        })
 }
 
 #[derive(Clone, Debug, From, Snafu, Deserialize, Serialize)]
@@ -125,20 +151,27 @@ where
         include_str!("../api/availability.toml"),
         options.extensions.clone(),
     )?;
+    let timeout = options.fetch_timeout;
+
     api.with_version("0.0.1".parse().unwrap())
-        .get("get_leaf", |req, state| {
+        .get("get_leaf", move |req, state| {
             async move {
                 let id = match req.opt_integer_param("height")? {
                     Some(height) => LeafId::Number(height),
                     None => LeafId::Hash(req.blob_param("hash")?),
                 };
-                state.get_leaf(id).await.context(FetchLeafSnafu {
-                    resource: id.to_string(),
-                })
+                state
+                    .get_leaf(id)
+                    .await
+                    .with_timeout(timeout)
+                    .await
+                    .context(FetchLeafSnafu {
+                        resource: id.to_string(),
+                    })
             }
             .boxed()
         })?
-        .stream("stream_leaves", |req, state| {
+        .stream("stream_leaves", move |req, state| {
             async move {
                 let height = req.integer_param("height")?;
                 state
@@ -150,7 +183,7 @@ where
             .try_flatten_stream()
             .boxed()
         })?
-        .get("get_header", |req, state| {
+        .get("get_header", move |req, state| {
             async move {
                 let id = if let Some(height) = req.opt_integer_param("height")? {
                     BlockId::Number(height)
@@ -162,6 +195,8 @@ where
                 Ok(state
                     .get_block(id)
                     .await
+                    .with_timeout(timeout)
+                    .await
                     .context(FetchBlockSnafu {
                         resource: id.to_string(),
                     })?
@@ -170,7 +205,7 @@ where
             }
             .boxed()
         })?
-        .stream("stream_headers", |req, state| {
+        .stream("stream_headers", move |req, state| {
             async move {
                 let height = req.integer_param("height")?;
                 Ok(state
@@ -188,7 +223,7 @@ where
             .try_flatten_stream()
             .boxed()
         })?
-        .get("get_block", |req, state| {
+        .get("get_block", move |req, state| {
             async move {
                 let id = if let Some(height) = req.opt_integer_param("height")? {
                     BlockId::Number(height)
@@ -197,13 +232,18 @@ where
                 } else {
                     BlockId::PayloadHash(req.blob_param("payload-hash")?)
                 };
-                state.get_block(id).await.context(FetchBlockSnafu {
-                    resource: id.to_string(),
-                })
+                state
+                    .get_block(id)
+                    .await
+                    .with_timeout(timeout)
+                    .await
+                    .context(FetchBlockSnafu {
+                        resource: id.to_string(),
+                    })
             }
             .boxed()
         })?
-        .stream("stream_blocks", |req, state| {
+        .stream("stream_blocks", move |req, state| {
             async move {
                 let height = req.integer_param("height")?;
                 Ok(state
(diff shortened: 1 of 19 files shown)


src/availability.rs
```diff
@@ -280,9 +333,9 @@ where
 mod test {
```
Ayiga, line 333: This test block requires the `"no-storage"` feature in order to build correctly.

```suggestion
#[cfg(all(test, feature = "no-storage"))]
mod test {
```

Ayiga: Careful here, it wouldn't let me select the `cfg` line for the `suggestion` for some reason.

src/data_source/fetching.rs
```diff
@@ -93,21 +253,36 @@ where
...
+        Builder::new(storage, provider)
     }
 
-    pub async fn with_retry_delay(
-        storage: S,
-        provider: P,
-        delay: Option<Duration>,
-    ) -> anyhow::Result<Self> {
+    /// Create a data source with local storage and a remote data availability provider.
+    pub async fn new(builder: Builder<Types, S, P>) -> anyhow::Result<Self> {
```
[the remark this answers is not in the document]
jbearer: Not 100% following...are you suggesting basically replace `FetchingDataSource::new` with `Builder::build` or something? I wanted to keep the implementation of this function in `FetchingDataSource` since the logic is specific to this type. But interface wise I agree...maybe we should make this function private and have `Builder::build` be the only way to construct a `FetchingDataSource`. WDYT?

src/availability.rs
jbearer: Oh yeah, we generally just require `--all-features` for building test code. Cuz even if we made it work, building without `--all-features` would probably not be what you want as it would silently skip a bunch of tests

src/data_source/fetching.rs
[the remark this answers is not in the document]
Ayiga: Essentially yes.  Following a Traditional `Builder` pattern the `build` call should do all of the setup work.  But at the same time the `FetchingDataSource::new` call really shouldn't depend on the `Builder`, as it implies it is unable to be built without a `Builder`, which essentially makes it not the `Builder` pattern.


As an experiment, I wrote the following types to try and work around it:

```rust
pub struct FetchingDataSourceConfig<Types, S, P>
where
    Types: NodeType,
{
    minor_scan_interval: Duration,
    major_scan_interval: usize,
    proactive_fetching: bool,
    fetcher: Fetcher<Types, S, P>,
}
```

Then the `FecthingDataSource::new` function would look something similar to:

```rust
pub async fn new(config: FetchingDataSourceConfig<Types, S, P>) -> anyhow::Result<Self> {
        let proactive_fetching = config.proactive_fetching;
        let minor_interval = config.minor_scan_interval;
        let major_interval = config.major_scan_interval;

        let fetcher = Arc::new(config.fetcher);
        let scanner = if proactive_fetching {
            Some(BackgroundTask::spawn(
                "proactive scanner",
                fetcher
                    .clone()
                    .proactive_scan(minor_interval, major_interval),
            ))
        } else {
            None
        };

        let mut ds = Self {
            fetcher,
            scanner,
            metrics: Default::default(),
        };

        if NodeDataSource::block_height(&ds).await? == 0 {
            // HotShot doesn't emit an event for the genesis block, so we need to manually ensure it
            // is present.
            ds.insert_genesis().await?;
        }

        Ok(ds)
    }
```

But we'd also need to do the same for the `Fetcher`, since it has the same dependency in it's `Fetcher::new` call.

```rust
pub struct FetcherConfig<Types, S, P>
where
    Types: NodeType,
{
    storage: S,
    provider: P,
    retry_delay: Option<Duration>,
    range_chunk_size: usize,

    _types: PhantomData<Types>,
}
```

Thus making the `Fetcher::new` call look like:

```
pub struct FetcherConfig<Types, S, P>
where
    Types: NodeType,
{
    storage: S,
    provider: P,
    retry_delay: Option<Duration>,
    range_chunk_size: usize,

    _types: PhantomData<Types>,
}
```

And finally resulting in a `Builder::build` function that looks like this:

```rust
pub async fn build(self) -> anyhow::Result<FetchingDataSource<Types, S, P>> {
        let fetcher = Fetcher::new(FetcherConfig {
            storage: self.storage,
            provider: self.provider,
            retry_delay: self.retry_delay,
            range_chunk_size: self.range_chunk_size,
            _types: Default::default(),
        })
        .await?;

        FetchingDataSource::new(FetchingDataSourceConfig {
            minor_scan_interval: self.minor_scan_interval,
            major_scan_interval: self.major_scan_interval,
            proactive_fetching: self.proactive_fetching,
            fetcher: fetcher,
        })
        .await
    }
```

This removes the `Builder` from being an argument to the `Fetcher::new`, and `FetchingDataSource::new` calls while still allowing the `Builder::build` to create a new instance of `FetchingDataSource`.  In addition it keeps the building portion of the `Builder::build` call all internalized as much as possible.  It also allows `Fetcher::new` and `FetchingDataSource::new` to be created without a `Builder`, should one choose to.

The down side is two new bespoke configuration types for each.  The alternative would be to keep adding more arguments to the `new` functions, but that makes any changes to their data type requirements more annoying to change at the function invocation site.  Either way the Call site would need to suffer, without the `Builder`.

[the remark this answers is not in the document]
Ayiga: Functionally there's no difference here.  Things just get moved around.  But a `Builder` as a concept remains `distinct`.  You *could* achieve the same affect by renaming the `new` functions, and keeping them scoped so that only the `Builder` can see them.

src/availability.rs
Ayiga: That makes sense.  It was just something that the `rust-analyzer` was complaining about.

src/availability.rs
```diff
@@ -49,6 +49,19 @@ pub struct Options {
     #[arg(long = "availability-api-path", env = "HOTSHOT_AVAILABILITY_API_PATH")]
     pub api_path: Option<PathBuf>,
 
+    /// Timeout for failing requests due to missing data.
+    ///
+    /// If data needed to respond to a request is missing, it can (in some cases) be fetched from an
+    /// external provider. This parameter controls how long the request handler will wait for
+    /// missing data to be fetched before giving up and failing the request.
+    #[arg(
```
nomaxg, line 57: Not for this PR, but I think that timeout logic should eventually be handled at the tide-disco level.

jbearer: Hmm. It is (actually the underlying http client (surf/tide) has a timeout) but this is different from a general request timeout. This is "we don't have the data and couldn't get it within a short time frame, so we're explicitly returning an error", and ultimately the difference is we return 404 instead of 408

src/data_source/fetching.rs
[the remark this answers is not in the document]
jbearer: Gotcha. That all makes sense. For the time being I don't think we gain much from being able to construct the `FetchingDataSource` without the builder via `new`, and as you pointed at there's a maintenance burden from introducing those new config types. So for now my inclination is to make the unnecessary constructors private, and we can improve the interface if/when we ever have a need to expose them.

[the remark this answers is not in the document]
jbearer: Done in [a663b2c](https://github.com/EspressoSystems/hotshot-query-service/pull/390/commits/a663b2c7dad400dcbeb80d6ceb2da11580f24b70)

src/data_source/fetching.rs
```diff
@@ -634,10 +814,51 @@ where
...
+            prev_height = block_height;
+
+            // Iterate over all blocks that we should have. Merely iterating over the `Fetch`es
+            // without awaiting them is enough to trigger active fetches of missing blocks, since
+            // we always trigger an active fetch when fetching by block number. Moreover, fetching
+            // the block is enough to trigger an active fetch of the corresponding leaf if it too is
+            // missing, so this one loop takes care of all resources that might be missing.
+            //
+            // The chunking behavior of `get_range` automatically ensures that, no matter how big
+            // the range is, we will release the read lock on storage every `chunk_size` items, so
```
nomaxg, line 847: It may be a good idea to have the proactive fetching range chunk be a different parameter, just in case we want to tune the how greedy the background task is on the lock.

src/task.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+/// A background task which is cancelled on [`Drop`]
+///
+/// This handle can be cloned; cloning it does not clone the underlying task. There may be many
+/// handles to the same background task, and the task will be cancelled when all handles are
+/// dropped.
+#[derive(Clone, Debug)]
+pub struct BackgroundTask {
+    inner: Option<Arc<BackgroundTaskInner>>,
```
nomaxg, line 35: Why is this optional?

nomaxg (approved):
LGTM, just one suggestion and a question

src/data_source/fetching.rs
jbearer: Good idea, will do

nomaxg (approved):

Issue #389, opened 2024-02-02
Add Docker remote host considerations for tests


Ayiga:
As it stands right now there are some potentially undocumented requirements for running the `tests` for `HotShot Query Service`.

The primary two that I am aware of at the moment are as follows:
- Docker is running on the local machine
- `psql` program for Postgres is installed

The second point is easily addressable once understood, but the first can be difficult.  It is ultimately due to an assumption of `"localhost"` being a way to connect to the spun up `postgres` instance.  For `Docker` instances running on remote machines, this **may not** be `"localhost"`.  As such, this assumption prevents us from being able to run this test on machines that use a remote `Docker Host`.

This issue proposes to address the inability to use a Remote Docker Host to run tests

PR #391, opened 2024-02-05
Add host to TmpDb struct

Ayiga:
The first barrier to being able to remotely target a Docker
container is the inability to target a host other than
localhost.

This commit adds host to the TmpDb struct, and utilizes it
to populate the generated configuration. This will allow for
the client that consumes the Config to be able to target
a different host other than "localhost".

Add derived hostname for docker host

The second barrier is to actually populate the now
configurable Hostname.  We require the ability to specify
the hostname of the remote docker instance.

In order to fulfill this requirement I've opted to specify
the remote docker hostname via an environment variable
named "DOCKER_HOSTNAME".  We don't want to use
"DOCKER_HOST" as it's already in use by `docker` itself
for specifying a different target.  In this case
"DOCKER_HOSTNAME" is expected to just be a string of
the hostname itself.  It should even be able to be an ip
address if one should desire it to be.

Fixes Add Docker remote host considerations for tests #389

### DIFF
--- a/src/data_source/storage/sql.rs
+++ b/src/data_source/storage/sql.rs
@@ -1343,6 +1343,7 @@ pub mod testing {
     use crate::testing::sleep;
     use portpicker::pick_unused_port;
     use std::{
+        env,
         process::{Command, Stdio},
         str,
         time::Duration,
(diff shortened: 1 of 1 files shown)


jbearer (approved):
LGTM

PR #401, opened 2024-02-08
Add `/node/sync-status` endpoint to get information about missing data

### LINKED ISSUE #362: Add metrics for missing data
Now that we have a system where data is allowed to be missing and might be fetched asynchronously, it is useful to know how much/which data is missing. We should add endpoints to get the counts of missing payloads and leaves in the availability APIs. We can use these to build automated alerts whenever data is missing.

jbearer:
This endpoint can be used to determine if the database is fully
synced and, if not, how much data is missing. This is useful for
alerts (@Ancient123) as well as for the block explorer to be able to inform the
user when aggregate queries may be inaccurate due to syncing in
progress (@Ayiga).

The query used to count missing data is not indexed or optimized,
but I believe it can be made efficient in the future:
* The `count(*)` to determine the number of leaves we have can be
  slow when the `leaf` table is large, because Postgres uses a full
  table scan to count. This could be made more efficient a number of
  ways, including:
  - Using estimated counts instead of exact counts
  - Storing in the database the latest count and using triggers to
    update it on inserts/updates
  - Inserting placeholder rows with NULL columns for leaves we don't
    have. This would ensure that there is a row for each possible leaf
    height, and the presence or absence of the leaf becomese a boolean
    property we can express on that row. Instead of counting over the
    entire table, we only need to count `WHERE data IS NULL` or similar,
    and we can use an index to efficiently process that filter.
* The `WHERE data IS NULL` filter for counting missing payloads is not
  indexed. We will add an index there after we test with a large enough
  database to be confident that Postgres is using the index and it is
  improving performance.

These optimizations are intentionally left as future work, for after we
have run this for some time and with large enough datasets that we can
really see the effects of various optimization strategies.

Of course, another approach to optimization would be to keep the count
of missing blocks and leaves in memory, and maintain it whenever we
`insert_block` or `insert_leaf` -- dead reckoning. This would probably
be the most efficient approach, but since this endpoint is designed to
be used for alerting and detecting low-level incidents like bugs in the
storage layer or failures in RDS, in many of the situations we care about,
the actual state of the database could end up diverging from our dead
reckoning estimate, and we might miss an incident. Therefore, I think
the endpoint is much more useful if it is running some meaningful
query over the actual data.

Closes #362

### DIFF
--- a/src/data_source/storage/sql.rs
+++ b/src/data_source/storage/sql.rs
@@ -19,7 +19,7 @@ use crate::{
         TransactionHash, TransactionIndex, UpdateAvailabilityData,
     },
     data_source::VersionedDataSource,
-    node::{NodeDataSource, UpdateNodeData},
+    node::{NodeDataSource, SyncStatus, UpdateNodeData},
     Header, Leaf, MissingSnafu, NotFoundSnafu, Payload, QueryError, QueryResult, SignatureKey,
 };
 use async_std::{
@@ -388,9 +388,7 @@ impl SqlStorage {
             kill: Some(kill),
         })
     }
-}
 
-impl SqlStorage {
     /// Access the transaction which is accumulating all uncommitted changes to the data source.
     pub async fn transaction(&mut self) -> QueryResult<Transaction<'_>> {
         if !self.tx_in_progress {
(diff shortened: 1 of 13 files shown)


src/data_source/storage/sql.rs
```diff
@@ -774,6 +772,62 @@ where
         let count: i64 = row.get(0);
         Ok(count as usize)
     }
+
+    async fn sync_status(&self) -> QueryResult<SyncStatus> {
```
nomaxg, line 776: Are we worried at all about the possibility of spamming our nodes with this endpoint? Looks expensive

nomaxg (approved):
LGTM, just a question about impact on the node if someone were to spam this endpoint. We could always consider "authenticating" with a hardcoded env secret

src/data_source/storage/sql.rs
jbearer: Yes. But it shouldn't be *that* expensive. It's not the only endpoint we have that currently triggers a full table scan. I think we can optimize this to make it less expensive (will do this after we have seen the behavior in a realistic setting with large tables). In the long run, we probably want to implement some type of API credit system, where endpoints like this would require more credit than endpoints that are easier to index

PR #406, opened 2024-02-12
Add Block Summaries Endpoints

Ayiga:
For the sake of the Block Explorer, as well as potential future utility we
find ourselves in need of a data structure that can describe aspects of
a Block without transmitting the full contents of a Block.  Currently, we
would like the ability to retrieve a Summary of a Block that indicates the
properties of a Block, as well as a summary of how many transactions
are contained within the Block.  In addition we find ourselves needing
to be able to grab a range of these Block Summaries for Pagination
and comparison purposes.

Add BlockSummaryQueryData type

The BlockSummaryQueryData defines the data layout of the Block
Summary endpoints.

Add From trait for BlockQuerySummaryData from BlockQueryData

Since the BlockQuerySummaryData will eventually be able to be
generated entirely from the BlockQueryData, it makes sense to be
able to actively convert to the Summary type given a Block itself.

Add get_block_summary endpoint

This endpoint can retrieve a single block_summary for a given
block height to indicate its position within the ledger.

Add get_block_summary_range endpoint

This endpoint allows us to retrieve a a range of block summaries
so we can grab a collection of them for direct reference.

### DIFF
--- a/api/availability.toml
+++ b/api/availability.toml
@@ -197,3 +197,20 @@ Get the stake table after `:height` leaves or after `:view_number`.
 
 Returns a map from staking public keys to amounts.
 """
+
+
+[route.get_block_summary]
+PATH = ["block/summary/:height"]
+":height" = "Integer"
+DOC = """
+Get the Block Summary for a block based on its position in the ledger.
+"""
+
+[route.get_block_summary_range]
+PATH = ["block/summaries/:from/:until"]
+":from" = "Integer"
+":until" = "Integer"
+DOC = """
+Get the Block Summary entries for blocks based on their position in the ledger,
+the blocks are taken starting from the given :from up until the given :until.
+"""
--- a/src/availability.rs
+++ b/src/availability.rs
@@ -331,14 +331,14 @@ where
                 let id: usize = req.integer_param("height")?;
 
                 state
-                    .get_block(BlockId::Number(id))
+                    .get_block(id)
                     .await
                     .with_timeout(timeout)
                     .await
                     .context(FetchBlockSnafu {
                         resource: id.to_string(),
                     })
-                    .map(Into::<BlockSummaryQueryData<Types>>::into)
+                    .map(BlockSummaryQueryData::from)
             }
             .boxed()
         })?
@@ -351,12 +351,12 @@ where
                     .get_block_range(from..until)
                     .await
                     .enumerate()
-                    .map(|(index, fetch)| {
-                        fetch.context(FetchBlockSnafu {
+                    .then(|(index, fetch)| async move {
+                        fetch.with_timeout(timeout).await.context(FetchBlockSnafu {
                             resource: (index + from).to_string(),
                         })
                     })
-                    .map(|result| result.map(|block| block.into()))
+                    .map(|result| result.map(BlockSummaryQueryData::from))
                     .try_collect()
                     .await?;
 
@@ -488,6 +488,30 @@ mod test {
                     .await
                     .unwrap(),
             );
+            let block_summary = client
+                .get(&format!("block/summary/{}", i))
+                .send()
+                .await
+                .unwrap();
+            assert_eq!(
+                BlockSummaryQueryData::<MockTypes>::from(block.clone()),
+                block_summary,
+            );
+            assert_eq!(block_summary.header(), block.header());
+            assert_eq!(block_summary.hash(), block.hash());
+            assert_eq!(block_summary.size(), block.size());
+            assert_eq!(
+                block_summary.num_transactions(),
+                block.num_transactions() as u64
+            );
+
+            let block_summaries: Vec<BlockSummaryQueryData<MockTypes>> = client
+                .get(&format!("block/summaries/{}/{}", 0, i))
+                .send()
+                .await
+                .unwrap();
+            assert_eq!(block_summaries.len() as u64, i);
+
             // We should be able to look up the block by payload hash. Note that for duplicate
             // payloads, these endpoints may return a different block with the same payload, which
             // is acceptable. Therefore, we don't check equivalence of the entire `BlockQueryData`
@@ -661,25 +685,8 @@ mod test {
                 block,
                 client.get(&format!("block/{}", i)).send().await.unwrap()
             );
-            let block_summary: BlockSummaryQueryData<MockTypes> = client
-                .get(&format!("block/summary/{}", i))
-                .send()
-                .await
-                .unwrap();
-
-            assert_eq!(block_summary.header(), block.header());
-            assert_eq!(block_summary.hash(), block.hash());
-            assert_eq!(block_summary.size(), block.size());
-            assert_eq!(block_summary.num_transactions(), 1);
 
             validate(&client, (i + 1) as u64).await;
-
-            let block_summaries: Vec<BlockSummaryQueryData<MockTypes>> = client
-                .get(&format!("block/summaries/{}/{}", i, i + 1))
-                .send()
-                .await
-                .unwrap();
-            assert_eq!(block_summaries.len(), 1);
         }
 
         network.shut_down().await;
--- a/src/availability.rs
+++ b/src/availability.rs
@@ -30,7 +30,7 @@ use crate::{api::load_api, Payload};
 use clap::Args;
 use cld::ClDuration;
 use derive_more::From;
-use futures::{FutureExt, StreamExt, TryFutureExt};
+use futures::{FutureExt, StreamExt, TryFutureExt, TryStreamExt};
 use hotshot_types::traits::node_implementation::NodeType;
 use serde::{Deserialize, Serialize};
 use snafu::{OptionExt, Snafu};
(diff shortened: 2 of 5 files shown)


api/availability.toml
```diff
@@ -197,3 +197,20 @@ Get the stake table after `:height` leaves or after `:view_number`.
 
 Returns a map from staking public keys to amounts.
 """
+
+
+[route.get_block_summary]
+PATH = ["block/summary/:height"]
```
jbearer, line 203: I think it makes sense to index these summaries the same way we index the full blocks, ie we would also want to add `block/summary/hash/:hash` and `block/summary/payload-hash/:payload-hash`. But if this is not blocking the block explorer, we can save it for future work.

src/availability.rs
```diff
@@ -325,6 +325,44 @@ where
...
+
+                state
+                    .get_block(BlockId::Number(id))
+                    .await
+                    .with_timeout(timeout)
+                    .await
+                    .context(FetchBlockSnafu {
+                        resource: id.to_string(),
+                    })
+                    .map(|block| Into::<BlockSummaryQueryData<Types>>::into(block))
```
jbearer: Nit: slightly more idiomatic to say `BlockSummaryQueryData::from(block)`. You might even be able to do `block.into()`, but maybe the compiler can't infer the type there

src/availability.rs
```diff
@@ -325,6 +325,44 @@ where
                     .unwrap())
             }
             .boxed()
+        })?
+        .get("get_block_summary", move |req, state| {
+            async move {
+                let id: usize = req.integer_param("height")?;
+
+                state
+                    .get_block(BlockId::Number(id))
```
jbearer, line 334: Nit: should be able to just use `id` on its own, as `get_block` is already generic over `T: Into<BlockId>`

src/availability.rs
```diff
@@ -325,6 +325,44 @@ where
...
+                    .await
+                    .with_timeout(timeout)
+                    .await
+                    .context(FetchBlockSnafu {
+                        resource: id.to_string(),
+                    })
+                    .map(Into::<BlockSummaryQueryData<Types>>::into)
```
jbearer, line 341: Nit: slightly more idiomatic to use `BlockSummaryQueryData::from`

src/availability.rs
```diff
@@ -325,6 +325,44 @@ where
...
+            async move {
+                let from: usize = req.integer_param("from")?;
+                let until: usize = req.integer_param("until")?;
+
+                let result: Vec<BlockSummaryQueryData<Types>> = state
+                    .get_block_range(from..until)
+                    .await
+                    .enumerate()
+                    .map(|(index, fetch)| {
+                        fetch.context(FetchBlockSnafu {
```
jbearer, line 355: I think to be consistent with the rest of the API, we should use `fetch.with_timeout` here (ie fail if we don't get the block after a short time, rather than failing immediately). This will require an `await` in the closure here, so you can use `.then` instead of `.map` to make the closure `async`

src/availability.rs
```diff
@@ -623,7 +661,25 @@ mod test {
                 block,
                 client.get(&format!("block/{}", i)).send().await.unwrap()
             );
+            let block_summary: BlockSummaryQueryData<MockTypes> = client
```
jbearer, line 664: For consistency I think we should move the new queries and assertions into `validate` with the rest of the assertions

src/availability/query_data.rs
```diff
@@ -429,3 +429,84 @@ pub(crate) fn payload_size<Types: NodeType>(payload: &Payload<Types>) -> u64 {
...
+
+    pub fn size(&self) -> u64 {
+        self.size
+    }
+
+    pub fn num_transactions(&self) -> u64 {
+        self.num_transactions
+    }
+
+    pub fn proposer_id(&self) -> SignatureKey<Types> {
```
jbearer, line 465: Nit: the equivalent getter on `LeafQueryData` is just called `proposer`, I don't have a strong preference which one we use but we should be consistent

src/availability/query_data.rs
```diff
@@ -429,3 +429,84 @@ pub(crate) fn payload_size<Types: NodeType>(payload: &Payload<Types>) -> u64 {
...
+        self.proposer_id.clone()
+    }
+}
+
+// Get the Proposer from the given header.  At the moment the proposer cannot
+// be determined from the header, however it is intended to in the future.
+// For now we will just generate one with dummy data.
+// TODO update this function to retrieve the proposer from the Header type once
+//      the Header type has been updated to include the proposer.
```
jbearer, line 474: Can you make a github issue for this TODO?

src/availability/query_data.rs
```diff
@@ -429,3 +429,84 @@ pub(crate) fn payload_size<Types: NodeType>(payload: &Payload<Types>) -> u64 {
...
+pub fn get_proposer<Types: NodeType>(_header: &Header<Types>) -> SignatureKey<Types> {
```
jbearer, line 475: This should probably be private, or `pub(crate)` if necessary. It's not really part of the interface of this crate

jbearer (commented):
*Wow* this is so much simpler than adding new SQL queries and fetching logic for the summary types. Love to see it. I had a few minor comments, mostly on Rust idiom stuff, once that is addressed this looks good 👍

src/availability/query_data.rs
```diff
@@ -429,3 +429,84 @@ pub(crate) fn payload_size<Types: NodeType>(payload: &Payload<Types>) -> u64 {
...
+// Since BlockSummaryQueryData can be derived entirely from BlockQueryData, we
+// implement the From trait to allow for a seamless conversion using rust
+// contentions.
+impl<Types: NodeType> From<BlockQueryData<Types>> for BlockSummaryQueryData<Types>
+where
+    Payload<Types>: QueryablePayload,
+{
+    fn from(value: BlockQueryData<Types>) -> Self {
+        let payload = value.payload();
+        let size = payload_size::<Types>(&payload.clone());
```
jbearer, line 501: This clone is unnecessary (surprised Clippy doesn't catch this). No point in cloning the reference just to take a reference to the clone. We should be able to just use `payload`

jbearer: But also we could just use `value.size()` here

src/availability/query_data.rs
```diff
@@ -429,3 +429,84 @@ pub(crate) fn payload_size<Types: NodeType>(payload: &Payload<Types>) -> u64 {
...
+        let num_transactions = payload.len(value.metadata());
```
jbearer, line 502: Could just use `value.len()` here

src/availability/query_data.rs
```diff
@@ -429,3 +429,84 @@ pub(crate) fn payload_size<Types: NodeType>(payload: &Payload<Types>) -> u64 {
...
+
+        BlockSummaryQueryData {
+            header: value.header().clone(),
```
jbearer, line 505: Since we have ownership of `value` here, we could move out of header to avoid a clone:
```suggestion
            header: value.header,
```

src/availability/query_data.rs
```diff
@@ -429,3 +429,84 @@ pub(crate) fn payload_size<Types: NodeType>(payload: &Payload<Types>) -> u64 {
...
+            hash: value.hash(),
+            size,
+            num_transactions: num_transactions as u64,
+            proposer_id: get_proposer::<Types>(value.header()),
```
jbearer, line 509: For consistency, I think we should add `num_transactions` and `proposer_id` to `BlockQueryData` as well. Really, `BlockQueryData` should just be
```rust
struct BlockQueryData<Types> {
  summary: BlockSummaryQueryData<Types>,
  payload: Payload<Types>,
}
```

But to keep the JSON responses flat, it might not be a bad idea to just copy the fields of `BlockSummaryQueryData` into `BlockQueryData`. Anyways, if we do that we can just use `value.num_transactions` and `value.proposer_id` here

src/availability.rs
Ayiga: I initially tried chaining a `.with_timeout` off of the `get_block_range` off of this call, but it doesn't seem to exist on the `get_block_range` call. My guess is because it returns a `Stream` rather than a `Future`. I **could** add the `timeout` in the `map` call, where the `context` call is made if that makes more sense.

Additionally because this is a `Stream` of values, I don't think I can actually call `then` on the result.

src/availability/query_data.rs
Ayiga: If we change this to just `value.header`, we end up with a partially borrowed reference for the next reference to `value.header` in the call to `get_proposer`.  The only way I have to resolve this at the moment is to add a `.clone()` to the `value.header` call as `BlockHeader` doesn't implement the `Copy` trait.

Unless you have a different suggestion?

Ayiga: I see if we implement the suggestion / feedback for the `BlockQueryData<Types>`, this concern *may* be resolved.

src/availability/query_data.rs
Ayiga: Making this change requires us to be able to populate these values in `BlockQueryData::new`, and outside.  These values are ultimately determined by the `payload` with the `trait` `QueryablePayload`, which ultimately means polluting these `impl` blocks with more instances of the `Payload<Types> : QueryablePayload` restriction.

My current attempts have me attempting to limit these restrictions to `BlockQueryData::new` and `BlockQueryData::gensis` which seems to work.  However, I also need to add it to the `sql.rs` file for `parse_block`, which ultimately spreads to `parse_payload`.

The addition to `BlockQueryData::new` also affects `src/data_source/update.rs:88` which seems like this is leaking too much.

jbearer: Ah yeah that trait bound will cascade everywhere

jbearer: It is actually fine to put that trait bound everywhere it's needed to make `new` work (e.g. `update.rs` would be fine). If it is too much work we could also save this for a follow-up

src/availability.rs
Ayiga: Ah I see.  That's a really weird `API` definition

jbearer (approved):
Awesome! I think we're good here, once the build is fixed

src/availability/query_data.rs
Ayiga: Done: [https://github.com/EspressoSystems/hotshot-query-service/issues/407](https://github.com/EspressoSystems/hotshot-query-service/issues/407)

api/availability.toml
Ayiga: Opting to save this for the future.

PR #428, opened 2024-02-20
Add endpoint for transaction pagination reverse order

Ayiga:
There is a need to be able to retrieve the various entries of
the HotShot Query Service Availability Endpoints as ranges of
data in addition to single entries.

This change adds an endpoint that allows us to retrieve
transactions in a paginated manner provided we have a start
block height and offset that allows us to address the specific
transaction entry, and a limit that allows us to specify how
many transaction entries we want.  These three pieces of
information will ultimately allow for us to retrieve as many
transactions as we can.  This can also allow for pagination
for transactions.

### DIFF
--- a/src/availability.rs
+++ b/src/availability.rs
@@ -35,7 +35,7 @@ use hotshot_types::traits::node_implementation::NodeType;
 use serde::{Deserialize, Serialize};
 use snafu::{OptionExt, Snafu};
 use std::{
-    cmp::max, fmt::Display, num::NonZeroUsize, ops::Range, path::PathBuf, str::FromStr,
+    cmp::min, fmt::Display, num::NonZeroUsize, ops::Range, path::PathBuf, str::FromStr,
     time::Duration,
 };
 use tide_disco::{api::ApiError, method::ReadState, Api, RequestError, StatusCode};
@@ -483,28 +483,43 @@ where
                 let limit: usize = req.integer_param("limit")?;
                 let batch_size = NonZeroUsize::new(10).unwrap();
 
-                let step1 = build_range_stream_rev(height, batch_size);
-                let step2 = step1.flat_map(|range| {
-                    expand_range_to_block_range_stream::<State, Types>(range, state, timeout)
-                });
+                let chunked_ranges = build_range_iterator_rev(height, batch_size);
+                let stream_of_chunked_blocks = futures::stream::iter(chunked_ranges)
+                    .then(move |range| async move {
+                        let stream = expand_range_to_block_range_stream::<State, Types>(
+                            range, state, timeout,
+                        )
+                        .await;
 
-                let step3 = step2.flat_map(|result| {
-                    futures::stream::iter(match result {
-                        Ok(block) => expand_block_data_to_transaction_vec(block)
+                        let vec_of_blocks = stream
+                            .collect::<Vec<Result<BlockQueryData<Types>, Error>>>()
+                            .await;
+
+                        // We need to reverse here, otherwise we're not in the correct order.
+                        vec_of_blocks.into_iter().rev()
+                    })
+                    .flat_map(|iter| futures::stream::iter(iter));
+
+                let chunked_transactions = stream_of_chunked_blocks.map(|result| match result {
+                    Ok(block) => {
+                        let transactions = expand_block_data_to_transaction_vec(block);
+                        transactions
                             .into_iter()
                             .rev()
-                            .map(Some)
-                            .collect(),
-
-                        _ => vec![None],
-                    })
+                            .map(Ok)
+                            .collect::<Vec<Result<Transaction<Types>, Error>>>()
+                    }
+                    Err(err) => vec![Err(err)],
                 });
 
-                let result = step3
+                let streamed_transactions =
+                    chunked_transactions.flat_map(|vec| futures::stream::iter(vec.into_iter()));
+
+                let result = streamed_transactions
                     .skip(offset)
                     .take(limit)
-                    .collect::<Vec<Option<Transaction<Types>>>>()
-                    .await;
+                    .try_collect::<Vec<Transaction<Types>>>()
+                    .await?;
 
                 Ok(result)
             }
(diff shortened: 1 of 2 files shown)


src/availability.rs
```diff
@@ -363,10 +366,107 @@ where
...
 
+// build_range_stream_rev is a helper function that builds a stream of ranges
+// to query for.  It construcs thes streams starting from the end going toward
+// 0.
+fn build_range_stream_rev(
+    end: usize,
+    chunk_size: NonZeroUsize,
+) -> impl Stream<Item = Range<usize>> {
+    let chunk_size = chunk_size.get();
+    let steps = end / chunk_size;
```
jbearer, line 415: I think we need to round up here. If `end` is 3 and `chunk_size` is 2, then `steps = 3 / 2 = 1`, and we only get one step `1..2`, missing block 0.

src/availability.rs
```diff
@@ -363,10 +366,107 @@ where
...
+
+    futures::stream::iter((0..steps).map(move |chunk| {
+        // We start at the back such that step 0 *should* have it's
+        // end being the overall desired end size.
+        let end = end - ((steps - chunk) * chunk_size);
+        let start = max(0, end - chunk_size);
```
jbearer, line 421: This `max` doesn't do anything because we're using unsigned integers (as we should be!). `end - chunk_size` won't be negative, it will just underflow, which is very bad

src/availability.rs
```diff
@@ -363,10 +366,107 @@ where
...
+
+        start..end
+    }))
```
jbearer, line 424: I think this whole thing would be much easier to follow and less prone to underflows and the like if we built the iterator going forward and then just used `.rev()` at the end.

src/availability.rs
```diff
@@ -363,10 +366,107 @@ where
...
+    timeout: Duration,
+) -> impl Stream<Item = Result<BlockQueryData<Types>, Error>> + '_
+where
+    State: 'static + Send + Sync + ReadState,
+    <State as ReadState>::State: Send + Sync + AvailabilityDataSource<Types>,
+    Payload<Types>: QueryablePayload,
+{
+    state
+        .get_block_range(range.clone())
+        .into_stream()
```
jbearer, line 441: We should just `await` here (and make this function `async`). No need to convert into a nested stream and then `flatten`

src/availability.rs
```diff
@@ -363,10 +366,107 @@ where
...
+
+// expand_block_data_to_transaction_vec is a helper function that ultimately
+// takes a BlockQueryData and expands it into a collection of Transactions
+fn expand_block_data_to_transaction_vec<Types: NodeType>(
+    block_data: BlockQueryData<Types>,
+) -> Vec<Transaction<Types>>
+where
+    Payload<Types>: QueryablePayload,
+{
+    let data = block_data.clone();
```
jbearer, line 461: This `clone` should not be necessary

src/availability.rs
```diff
@@ -363,10 +366,107 @@ where
...
+        })?
+        .get("get_transaction_range_rev", move |req, state| {
+            async move {
+                let height: usize = req.integer_param("height")?;
+                let offset: usize = req.integer_param("offset")?;
+                let limit: usize = req.integer_param("limit")?;
+                let batch_size = NonZeroUsize::new(10).unwrap();
+
+                let step1 = build_range_stream_rev(height, batch_size);
+                let step2 = step1.flat_map(|range| {
```
jbearer, line 378: Please use more descriptive names (e.g. `let chunks =`, `let blocks = `, `let transactions = `

src/availability.rs
```diff
@@ -363,10 +366,107 @@ where
...
+
+                let step3 = step2.flat_map(|result| {
+                    futures::stream::iter(match result {
+                        Ok(block) => expand_block_data_to_transaction_vec(block)
+                            .into_iter()
+                            .rev()
+                            .map(Some)
+                            .collect(),
+
+                        _ => vec![None],
```
jbearer, line 390: I think I have convinced myself that we should just fail the whole endpoint if any blocks are missing, rather than inserting placeholders. The placeholders make it inconsistent with other endpoints in the availability API, plus, since we are querying for the blocks by number, we will always actively fetch them, and thus missing blocks should be resolved very quickly (within seconds), so I'm not sure the placeholders are really doing much besides complicating things.

If we really want the placeholders for the UI in the future, we could create a separate endpoint in a dedicated block explorer server

src/availability.rs
```diff
@@ -363,10 +366,107 @@ where
...
+        .get("get_transaction_range_rev", move |req, state| {
+            async move {
+                let height: usize = req.integer_param("height")?;
+                let offset: usize = req.integer_param("offset")?;
+                let limit: usize = req.integer_param("limit")?;
+                let batch_size = NonZeroUsize::new(10).unwrap();
+
+                let step1 = build_range_stream_rev(height, batch_size);
+                let step2 = step1.flat_map(|range| {
+                    expand_range_to_block_range_stream::<State, Types>(range, state, timeout)
```
jbearer, line 379: I think we need to reverse each one of these sub-ranges, since the chunks are in reverse order but within each chunk we are pulling the blocks in forward order

src/availability.rs
Ayiga: On the other hand, we **could** end up with the possibility of pulling more blocks than is necessary.  But we can always cap it with a `min` with the `end` in order to avoid such a scenario... So perhaps this is fine.

src/availability.rs
Ayiga: I do believe you are correct

src/availability.rs
```diff
@@ -404,29 +419,31 @@ where
...
 
-    futures::stream::iter((0..steps).map(move |chunk| {
-        // We start at the back such that step 0 *should* have it's
-        // end being the overall desired end size.
-        let end = end - ((steps - chunk) * chunk_size);
-        let start = max(0, end - chunk_size);
+    (0..steps)
+        .map(move |chunk| {
+            // We start at the back such that step 0 *should* have it's
+            // end being the overall desired end size.
```
jbearer, line 435: Not sure this comment is relevant anymore, with the forward iteration

jbearer (approved):
Looks good, once clippy is happy

src/availability.rs
```diff
@@ -472,10 +475,117 @@ where
...
+                    .await?;
+
+                Ok(result)
+            }
+            .boxed()
         })?;
     Ok(api)
 }
 
+// build_range_iterator_rev is a helper function that builds iterator of ranges
```
jparr721, line 531: Not sure if you meant for this, but using `//` instead of `///` results in no documentation showing up when applying `hover` within an editor. To fix this, you'll just need to change this comment (and the other functions below) to use `///` and it'll give hover support for the function.

jparr721 (commented):
Howdy! I'm the _new guy_. Feel free to disregard this comment, I was just browsing around and noticed this.

src/availability.rs
Ayiga: No worries @jparr721! I'm new to `rust` so I appreciate the insight!

Ayiga:
This approach has been rethought, and this PR is no longer relevant. Closing.

PR #429, opened 2024-02-20
add range endpoints for leaf, block, payload and header

imabdulbasit:


src/availability.rs
```diff
@@ -171,6 +171,26 @@ where
...
+                state
+                    .get_leaf_range(from..until)
+                    .await
+                    .enumerate()
+                    .then(|(index, fetch)| async move {
+                        fetch.with_timeout(timeout).await.context(FetchLeafSnafu {
+                            resource: (index + from).to_string(),
+                        })
+                    })
+                    .map(|result| result.map(LeafQueryData::from))
```
jbearer, line 188: Is there an error if we don't have this `LeafQueryData::from`? I thought `get_leaf_range` is already returning a stream of `LeafQueryData`.

src/availability.rs
```diff
@@ -243,6 +283,26 @@ where
...
+                state
+                    .get_block_range(from..until)
+                    .await
+                    .enumerate()
+                    .then(|(index, fetch)| async move {
+                        fetch.with_timeout(timeout).await.context(FetchBlockSnafu {
+                            resource: (index + from).to_string(),
+                        })
+                    })
+                    .map(|result| result.map(BlockQueryData::from))
```
jbearer, line 300: same question here

src/availability.rs
```diff
@@ -275,6 +335,26 @@ where
...
+                state
+                    .get_payload_range(from..until)
+                    .await
+                    .enumerate()
+                    .then(|(index, fetch)| async move {
+                        fetch.with_timeout(timeout).await.context(FetchBlockSnafu {
+                            resource: (index + from).to_string(),
+                        })
+                    })
+                    .map(|result| result.map(PayloadQueryData::from))
```
jbearer, line 352: and here

src/availability.rs
imabdulbasit: No error.

imabdulbasit:
force pushed the main branch mistakenly. creating PR again

Issue #409, opened 2024-02-12
Add range endpoints for block, payload, leaf, and header


jbearer:
These should be consistent with the range endpoints we just added for block summaries

PR #430, opened 2024-02-20
add range endpoints for leaf, block, payload and header

imabdulbasit:
addresses https://github.com/EspressoSystems/hotshot-query-service/issues/409

### DIFF
--- a/src/availability.rs
+++ b/src/availability.rs
@@ -171,6 +171,25 @@ where
             }
             .boxed()
         })?
+        .get("get_leaf_range", move |req, state| {
+            async move {
+                let from = req.integer_param::<_, usize>("from")?;
+                let until = req.integer_param("until")?;
+
+                state
+                    .get_leaf_range(from..until)
+                    .await
+                    .enumerate()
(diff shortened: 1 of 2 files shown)


src/availability.rs
```diff
@@ -579,6 +656,38 @@ mod test {
                         .hash()
                 );
             }
+
+            let block_range: Vec<BlockQueryData<MockTypes>> = client
```
imabdulbasit, line 660: Should I add tests to compare whole range or is that not neccessary?

jbearer: Compare against what? Fetching each block in the range individually? It would be good but would slow the test down a lot, and we already have tasks for the underlying `get_*_range` functions in `data_source.rs` (`test_range`). Maybe just spot check the most recent block in each range?

jbearer (approved):

src/availability.rs
imabdulbasit: I was thinking of collecting the individual blocks already fetched and comparing them, but since we are not doing that for all the blocks, this approach would fail.

PR #431, opened 2024-02-21
Status endpoint for elapsed time since last block

imabdulbasit:
addresses: https://github.com/EspressoSystems/hotshot-query-service/issues/411

Changes:

- Added status /time_since_last_decide endpoint to retrieve time since last block commit.
- Updated hotshot crates to version 0.5.15.


jbearer (approved):
LGTM

Issue #376, opened 2024-01-23
Implement pruning for old blocks


jbearer:
The current query service is archival: it keeps everything. We can easily support other use cases just by (optionally) pruning old blocks. For example, DA committee nodes are expected to persist and provide blocks for some time, but not forever.

jbearer:
Discussed today, we need a slightly more nuanced pruning policy to account for spiky utilization. Our goal is to retain data, for, e.g., 7 days under average load but at least 1 day under maximum load, while still having a reasonable bound on the total storage requirement. To that end, our policy should take two parameters:
* Target retention
* Minimum retention

If pruning is enabled, all data is pruned automatically after the target retention period. In addition, if we run out of storage, we are allowed to prune the oldest data that is older than minimum retention. If we run out of storage and all persisted data is younger than the minimum retention, it is an error, and we stop storing new data: node operators must provision enough storage to prevent this from happening.

We will need to somehow detect when a write fails due to being out of storage.

Both of these parameters are specified in wall clock time, not block number, since the thing we really care about here is we have enough wall clock time to retrieve and archive data before it is pruned.

jbearer:
We should implement two metrics/alerts with this:
* Oldest unpruned object. This is very important because it tells us directly how long we are keeping data before deleting it in current conditions. If this is two low, we need to take action because it means we don't have much time to persist data before it is pruned, in the event that something breaks with the archival nodes
* Storage utilization (or what the node things it is) as a factor of maximum utilization. If this gets down close to one, it means the node is close to failing to accept new data

jbearer:
Another tricky thing here: if we just delete data when pruning it, the fetcher will notice that data is missing and try to retrieve it and store it again. So the fetcher also needs to know about pruning, and avoid fetching data if it is out of the retention window.

jbearer:
Recording a discussion I had with @imabdulbasit on the minimum retention policy, and the question of how to determine when we are running out of storage and when the more aggressive pruning should kick in. Postgres has a function to [get the size of the database](https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADMIN-DBSIZE), but this does not tell us how much space is *left*. I think we will just leave this up to the node operator, so we will need an additional configuration parameter for the available space. If the size of the database exceeds this parameter, then we start pruning beyond the target retention.

[...]

PR #438, opened 2024-02-26
Data Pruning for SQL storage

imabdulbasit:
This PR introduces SQL data pruning to address https://github.com/EspressoSystems/hotshot-query-service/issues/376

includes pruning_threshold and max_usage to config to prune data older than threshold until DB size is < max usage.


src/data_source/storage/sql.rs
```diff
@@ -408,6 +425,143 @@ impl SqlStorage {
...
+                )
+                .await?;
+
+            if let Some(height) = minimum_retention_height {
+                let mut current_height =
+                    self.get_minimum_height().await?.ok_or(QueryError::Error {
+                        message: "Minimum height not found".to_string(),
+                    })?;
+
+                while (usage as f64 / cfg.pruning_threshold() as f64) > 0.7
```
jbearer, line 536: Should we make this 70% configurable?

src/testing/mocks.rs
```diff
@@ -37,6 +39,12 @@ pub fn mock_transaction(payload: Vec<u8>) -> MockTransaction {
     TestTransaction(payload)
 }
 
+impl QueryableHeader for MockHeader {
+    fn timestamp(&self) -> u64 {
+        random::<u64>() % (Utc::now().timestamp() as u64)
```
jbearer, line 44: Why `random`?

src/data_source/storage/pruning.rs
```diff
@@ -0,0 +1,94 @@
+use async_trait::async_trait;
+use std::error::Error;
+use std::fmt::Debug;
+#[async_trait]
+pub trait PruneStorage: PrunerConfig {
+    type Error: Error + Debug + Send + Sync + 'static;
+
+    async fn get_minimum_height(&self) -> Result<Option<u64>, Self::Error> {
```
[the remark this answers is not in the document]
imabdulbasit: get_minimum_height is a helper function to prune oldest data first. I have moved it out of this trait.

src/data_source/storage/sql.rs
```diff
@@ -408,6 +425,143 @@ impl SqlStorage {
...
+            }
+
+            self.pruned_height = Some(height);
+        }
+
+        let mut usage = self.get_disk_usage().await?;
+
+        // Prune data exceeding minimum retention in batches  starting from minimum height.
+        // Stop when disk usage is 70% of the pruning threshold.
+        if cfg.pruning_threshold() > 0 && usage > cfg.pruning_threshold() {
```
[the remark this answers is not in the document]
imabdulbasit: Changed to Option and added a validate function to keep it > 0 to avoid div by 0

src/testing/mocks.rs
imabdulbasit: changed to Utc::now() 
 was trying to insert mock data with range of timestamps into db for testing

src/data_source/storage/pruning.rs
```diff
@@ -0,0 +1,154 @@
...
+use std::error::Error;
+use std::fmt::Debug;
+use std::time::Duration;
+
+#[derive(Clone, Debug)]
+pub struct PrunerCfg {
+    /// Disk space threshold (in bytes).
+    ///
+    /// If the disk usage exceeds this threshold, pruning of data starts from
+    /// the oldest data and continues until the disk usage falls below `MAX_USAGE`.
```
jbearer, line 12: Or until the oldest data is younger than `minimum_retention`

src/data_source/storage/pruning.rs
```diff
@@ -0,0 +1,154 @@
...
+    pruning_threshold: Option<u64>,
+
+    /// Minimum data retention period (in seconds).
+    ///
+    /// Data younger than this is never pruned, regardless of disk usage.
+    minimum_retention: Duration,
+
+    /// Target data retention period (in seconds).
```
jbearer, line 20:
```suggestion
    /// Target data retention period.
```

src/data_source/storage/pruning.rs
```diff
@@ -0,0 +1,154 @@
...
+
+#[derive(Clone, Debug)]
+pub struct PrunerCfg {
+    /// Disk space threshold (in bytes).
+    ///
+    /// If the disk usage exceeds this threshold, pruning of data starts from
+    /// the oldest data and continues until the disk usage falls below `MAX_USAGE`.
+    pruning_threshold: Option<u64>,
+
+    /// Minimum data retention period (in seconds).
```
jbearer, line 15:
```suggestion
    /// Minimum data retention period.
```

src/data_source/storage/sql.rs
```diff
@@ -408,6 +426,135 @@ impl SqlStorage {
...
+                                tracing::info!("Pruned data up to height {}", height);
+                            }
+                            self.pruned_height = Some(height);
+                        }
+                    }
+                }
+                // Vacuum the database to reclaim space.
+                // Note: VACUUM FULL is not used as it requires an exclusive lock on the tables, which can
+                // cause downtime for the query service.
+                self.query_opt_static("VACUUM;").await?;
```
jbearer, line 547: Since `VACUUM` doesn't return anything, you may have to use the low-level `self.client.batch_execute("VACUUM")` here

src/data_source/storage/sql.rs
```diff
@@ -1762,6 +1915,25 @@ mod test {
         connect(true, migrations).await.unwrap();
     }
 
+    #[async_std::test]
+    async fn test_current_migrations() {
```
jbearer: I believe `test_migrations` already tests whatever is in the `migrations` directory

src/data_source/storage/sql.rs
```diff
@@ -1781,4 +1953,105 @@ mod test {
...
+
+    #[async_std::test]
+    async fn test_target_period_pruning() {
+        setup_test();
+
+        let db = TmpDb::init().await;
+        let port = db.port();
+        let host = &db.host();
+
+        let migrations = include_migrations!("$CARGO_MANIFEST_DIR/migrations").collect::<Vec<_>>();
```
jbearer, line 1946: You shouldn't need `include_migrations!` here, whatever is in the `migrations` directory is executed by default

src/data_source/storage/sql.rs
```diff
@@ -1781,4 +1953,105 @@ mod test {
...
+        }
+
+        // Set pruner config to default which has minimum retention set to 1 day
+        storage.set_pruning_config(PrunerCfg::new());
+        sleep(Duration::from_secs(2)).await;
+        // No data will be pruned
+        storage.prune().await.unwrap();
+        // Pruned height should be none
+        assert!(storage.pruned_height().is_none());
+
```
jbearer, line 1970: Maybe good to sanity check that all the data is still retrievable

src/data_source/storage/sql.rs
```diff
@@ -1781,4 +1953,105 @@ mod test {
...
+        // Set pruner config to target retention set to 1s
+        storage.set_pruning_config(PrunerCfg::new().with_target_retention(Duration::from_secs(1)));
+        // All of the data is now atleast one sec old
+        storage.prune().await.unwrap();
+
+        // Pruned height should be some
+        // the table should now contain only 1 row as data with height < target height is pruned
```
jbearer, line 1977: What is the 1 row? I thought all data was older than 1s by this point?

src/data_source/storage/sql.rs
```diff
@@ -1781,4 +1953,105 @@ mod test {
...
+        assert!(storage.pruned_height().is_some());
+
+        let count = storage
+            .query_one_static("select count(*) as count from header")
+            .await
+            .unwrap()
+            .get::<_, i64>("count");
+        assert_eq!(count, 1);
+    }
```
jbearer, line 1986: Assert that database disk usage has decreased?

jbearer (commented):
Looking good. Note we don't have any stability commitment right now, and we're not going to be updating any deployed code with these changes; we'll just redeploy from scratch. So feel free to incorporate all the schema changes into the `init_schema` migration for clarity. This upcoming testnet is probably the last time we'll have the option to do this.

src/data_source/fetching.rs
```diff
@@ -119,6 +122,8 @@ use leaf::LeafFetcher;
 use transaction::TransactionRequest;
 use vid::{VidCommonFetcher, VidCommonRequest};
 
+static PRUNED_HEIGHT: AtomicU64 = AtomicU64::new(0);
```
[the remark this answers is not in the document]
imabdulbasit: There was some unexpected behaviour in the test but actually that was because of  multiple proactive scan tasks

src/data_source/storage/sql.rs
imabdulbasit: changed the check when deleting rows to make it inclusive. Count is number of rows and  == 0

src/data_source/storage/sql.rs
imabdulbasit: already have that in different test but can add here too

src/fetching/provider/query_service.rs
```diff
@@ -677,6 +679,62 @@ mod test {
         );
     }
 
+    #[async_std::test]
+    async fn test_pruning() {
```
jbearer, line 683: I don't think this test is really specific to fetching at all, unlike the other tests in this file. Consider moving this to `data_source/sql.rs`? Really it should be a generic test, but I guess it won't work for the file system since we haven't implemented pruning there

src/data_source/storage/pruning.rs
```diff
@@ -0,0 +1,155 @@
...
+}
+
+#[async_trait]
+pub trait PruneStorage: PrunerConfig {
+    type Error: Error + Debug + Send + Sync + 'static;
+
+    async fn get_disk_usage(&self) -> Result<u64, Self::Error> {
+        Ok(0)
+    }
+    async fn get_height_by_timestamp(&self, _timestamp: i64) -> Result<Option<u64>, Self::Error> {
```
jbearer, line 47: I think this function is only used as an internal helper function in the SQL pruning implementation, and never generically via this trait? If so, we could consider removing it from the trait and just making it a private method of `SqlStorage`

src/data_source/storage/sql.rs
```diff
@@ -326,6 +340,8 @@ pub struct SqlStorage {
     client: Arc<Client>,
     tx_in_progress: bool,
     kill: Option<oneshot::Sender<()>>,
+    pruner_cfg: Option<PrunerCfg>,
+    pruned_height: Option<u64>,
```
jbearer, line 344: Store this in the database so we don't reset on restart?

src/data_source/storage/sql.rs
```diff
@@ -408,6 +428,142 @@ impl SqlStorage {
...
+                // Vacuum the database to reclaim space.
+                // Note: VACUUM FULL is not used as it requires an exclusive lock on the tables, which can
+                // cause downtime for the query service.
+                self.client
+                    .batch_execute("VACUUM")
+                    .await
+                    .map_err(postgres_err)?;
+            }
+            None => {
+                tracing::info!("database is empty, nothing to prune");
```
jbearer, line 559: Nit: move this up with an early return to reduce nesting, e.g.

```rust
let Some(minimum_height) = self.get_minimum_height().await? else {
    tracing::info!("database is empty, nothing to prune");
    return Ok(());
}
```

jbearer (commented):
I didn't see a test that data younger than the minimum retention is kept even if usage exceeds the threshold. Did I miss it or can you add it?

src/fetching/provider/query_service.rs
imabdulbasit: deleted this test. There are separate tests for pruning in sql.rs

src/data_source/storage/pruning.rs
```diff
@@ -0,0 +1,159 @@
...
+    }
+}
+
+#[async_trait]
+pub trait PrunedHeightStorage {
+    type Error: Error + Debug + Send + Sync + 'static;
+    async fn save_pruned_height(&mut self, _height: u64) -> Result<(), Self::Error> {
+        Ok(())
+    }
+    async fn load_pruned_height(&mut self) -> Result<Option<u64>, Self::Error> {
```
jbearer, line 38: Why `&mut` for a load?

src/data_source/fetching.rs
```diff
@@ -929,13 +1004,15 @@ where
         let mut prev_height = 0;
 
         for i in 0.. {
+            let minimum_block_height =
+                self.storage.read().await.pruned_height.unwrap_or(0) as usize;
             // Get the block height; we will look for any missing blocks up to `block_height`.
             let block_height = { self.storage.read().await.height } as usize;
```
jbearer, line 1010: We should use the same lock on storage to read the minimum block height and block height. E.g.
```rust
let (minimum_block_height, block_height) = {
  let storage = self.stoage.read().await;
  ...
};
```

src/data_source/storage/pruning.rs
```diff
@@ -0,0 +1,159 @@
...
+pub trait PruneStorage: PrunerConfig + PrunedHeightStorage {
+    async fn get_disk_usage(&self) -> Result<u64, Self::Error> {
+        Ok(0)
+    }
+
+    async fn prune(&mut self) -> Result<Option<u64>, Self::Error> {
+        Ok(None)
+    }
+
+    fn pruned_height(&self) -> Option<u64> {
```
jbearer, line 27: I don't think we use this anymore, do we? Seems like we are always using `load_pruned_height`, the pruned height stored in `NotifyStorage`, or the return value of `prune`.

src/data_source/storage/sql.rs
jbearer, line 344: Maybe we can get rid of this since we are storing the same thing at a higher level in `NotifyStorage`

src/data_source/storage/sql.rs
```diff
@@ -408,6 +428,157 @@ impl SqlStorage {
...
+    }
+}
+
+#[async_trait]
+impl PrunedHeightStorage for SqlStorage {
+    type Error = QueryError;
+    async fn save_pruned_height(&mut self, height: u64) -> Result<(), Self::Error> {
+        // id is set to 1 so that there is only one row in the table.
+        // height is updated if the row already exists.
+        self.query_opt(
```
jbearer, line 472: Can you use `self.upsert` here? It has a bunch of retry logic and error handling stuff baked in

jbearer (commented):
Looks good. Just a few more minor comments

src/data_source/storage/sql.rs
```diff
@@ -469,21 +467,29 @@ impl PrunedHeightStorage for SqlStorage {
...
+            .await?
+            .upsert(
+                "pruned_height",
+                ["id", "last_height"],
+                ["id"],
+                [[sql_param(&(1_i32)), sql_param(&(height as i64))]],
+            )
+            .await?;
+
+        self.commit().await.map_err(|err| QueryError::Error {
```
jbearer, line 480: Commit should probably be called at the higher level, as we do elsewhere. E.g. in the pruner task it could be like
* prune
* save pruned height
* commit

PR #445, opened 2024-02-28
Cleanup docker containers on termination

tbro:
The `--rm` switch tells docker to cleanup containers on termination. Without this an explicit `remove` is required after `stop` to clean up the container and a `prune` to remove arbitrary volumes left behind. While these are not difficult actions to perform, we are often unaware that we need to do them.

I've also changed `kill` to `stop` because `stop` in the drop fn because `stop` attempts to shutdown the container gracefully.

### DIFF
--- a/src/data_source/storage/sql.rs
+++ b/src/data_source/storage/sql.rs
@@ -1450,6 +1450,7 @@ pub mod testing {
 
             let output = Command::new("docker")
                 .arg("run")
+                .arg("--rm")
                 .arg("-d")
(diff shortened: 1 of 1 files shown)


jbearer (approved):

sveitser:
Since contrainers are now removed on termination why not also remove the explicit `docker rm`?

https://github.com/EspressoSystems/hotshot-query-service/pull/445/files#diff-cb1e794a808d62e9693c1daa28a4a01d2395ad907f5f72052bef92b8707436fdR1535-R1545

PR #457, opened 2024-03-05
Implement timestamp queries; drop proposer queries

jbearer:
Now that we have access to the header timestamp in the query service (needed for pruning) we can implement timestamp queries, moving this functionality from extensions in the sequencer into the core query service.

Conversely, since the proposer address has been moved into the block header and integrated with the Espresso fee system, it no longer makes sense to have proposer endpoints in the core query service. These can be implemented as extensions that rely on the specific structure of the sequencer header, if they are needed at all.

### DIFF
--- a/src/availability.rs
+++ b/src/availability.rs
@@ -487,6 +487,7 @@ mod test {
             mocks::{mock_transaction, MockHeader, MockPayload, MockTypes},
             setup_test,
         },
+        types::HeightIndexed,
         Error, Header,
     };
     use async_std::{sync::RwLock, task::spawn};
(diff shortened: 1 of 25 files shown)


src/availability.rs
```diff
@@ -487,6 +487,7 @@ mod test {
             mocks::{mock_transaction, MockHeader, MockPayload, MockTypes},
             setup_test,
         },
+        types::HeightIndexed,
```
Ayiga, line 490: This isn't used in this module as far as I can tell.

Ayiga: Ah this is a `trait`, I see.

jbearer: It's used for the `.height()` method of leaves, blocks, etc

Ayiga (approved):

PR #473, opened 2024-03-12
 State storage and queries

imabdulbasit:
- Stores merkle tree updates on every Decide event
- Query for getting the MerkleTree proof path


src/merklized_state/data_source.rs
```diff
@@ -0,0 +1,152 @@
...
+
+// This trait defines methods that a data source should implement
+// It enables retrieval of the membership path for a leaf node, which can be used to reconstruct the Merkle tree state.
+#[async_trait]
+pub trait MerklizedStateDataSource<Types>
+where
+    Types: NodeType,
+{
+    type Error: std::error::Error + std::fmt::Debug + Send + Sync + 'static;
+    async fn get_path<
```
jbearer, line 48: I think this is a bit messy and error prone. It makes it easy to accidentally mix different types of state in the same table, for example.

What I was imagining was that this _trait_ would be parameterized on the type of state being stored, and that in turn would tell you the types of the other generic parameter, the names of the tables and columns involved, etc. So e.g. we might have `trait MerklizedStateStorage<Types: NodeType, State: MerklizedState>`, and then add associated types to `MerklizedState` for `E` and `T` (I think it already has `I` and `A`).

This means this trait might be implemented twice in the sequencer, e.g. `impl MerklizedStateDataSource<SeqTypes, FeeMerkleTree> for SqlStorage` and `impl MerklizedStateDataSource<SeqTypes, BlocksMerkleTree> for SqlStorage`. This is fine IMO because we already have a customization point for updates (`UpdateStateData`).

src/merklized_state/data_source.rs
```diff
@@ -0,0 +1,152 @@
...
+        T: NodeValue + Send + CanonicalDeserialize,
+    >(
+        &self,
+        _state_type: &'static str,
+        _tree_height: usize,
+        _header_state_commitment_field: &'static str,
+        _snapshot: Snapshot<Types>,
+        _key: Value,
+    ) -> Result<MerklePath<E, I, T>, Self::Error> {
+        Ok(vec![MerkleNode::Empty])
```
jbearer, line 61: I don't think we should have a default impl for this function, and it certainly shouldn't silently succeed with a random merkle path

src/merklized_state/data_source.rs
```diff
@@ -0,0 +1,152 @@
...
+use std::{cmp::Ordering, sync::Arc};
+
+// This trait defines methods that a data source should implement
+// It enables retrieval of the membership path for a leaf node, which can be used to reconstruct the Merkle tree state.
+#[async_trait]
+pub trait MerklizedStateDataSource<Types>
+where
+    Types: NodeType,
+{
+    type Error: std::error::Error + std::fmt::Debug + Send + Sync + 'static;
```
jbearer, line 47: Trying to move away from associated error types as I've found them to be an unnecessary maintenance burden. I think it would be better to just have `get_path` return a `QueryResult`

src/merklized_state/data_source.rs
```diff
@@ -0,0 +1,152 @@
...
+        _snapshot: Snapshot<Types>,
+        _key: Value,
+    ) -> Result<MerklePath<E, I, T>, Self::Error> {
+        Ok(vec![MerkleNode::Empty])
+    }
+}
+
+// This trait defines methods for updating the storage with the merkle tree state.
+#[async_trait]
+pub trait UpdateStateData: Send + Sync {
```
jbearer, line 67: Same here, this should be generic on `State: MerklizedState` and the relevant parameters should be removed from `insert_merkle_nodes`.

src/merklized_state/data_source.rs
```diff
@@ -0,0 +1,152 @@
...
+}
+
+// This trait should be implemented by types that represent validated states.
+// It enables services using the module to implement update functionality.
+// Some services may manage multiple Merkle trees within their validated state,
+// requiring updates for all state types with each event.
+// Therefore, the `insert_merkle_nodes` method of `UpdateStateData` should be called
+// for all state types within the validated state.
+#[async_trait]
+pub trait UpdateStateStorage<Types: NodeType> {
```
jbearer, line 92: If we end up taking my suggestion above (about making the data source generic over the state type) I think it would be good to parameterize this trait on the data source being updated, e.g. `trait UpdateStateStorage<Types: NodeType, State: MerklizedState, DataSource: MerklizedStateDataSource<Types, State>>`. The reason is that this allows the caller to specify what constraints they require of the data source. E.g. in the sequencer we know we need `DataSource: UpdateStateData<FeeMerkleTree> + UpdateStateData<BlockMerkleTree>`. Then we can _remove_ the `UpdteStateData` constraint from `UpdateDataSource`, which in turn allows us to remove the state trait impls from data sources that don't support them, like file system, while still implementing `UpdateDataSource`. This would prevent us from even compiling code that tried to use the file system storage with a `ValidatedState` type that required support for merklized storage.

Meanwhile, for the mock validated state type, we could implement `UpdateStateStorage` with _no additional constraints_, since it doesn't have any merklized state to store. Then it could be used with any data source.

I think this would be a nice cleanup of the interfaces, lmk what you think.

src/merklized_state/data_source.rs
```diff
@@ -0,0 +1,152 @@
...
+    async fn update_storage(
+        &self,
+        storage: &mut impl UpdateStateData,
+        leaf: &Leaf<Types>,
+        delta: Arc<<<Types as NodeType>::ValidatedState as ValidatedState<Types>>::Delta>,
+    ) -> anyhow::Result<()>;
+}
+
+type MerkleCommitment<Types> = Commitment<Leaf<Types>>;
```
jbearer, line 101: Why is a merkle commitment a commitment to a leaf?

src/data_source/storage/sql.rs
```diff
@@ -1170,6 +1192,401 @@ where
...
+        T: NodeValue + Send + Sync,
+    >(
+        &mut self,
+        name: &'static str,
+        proof: Proof,
+        traversal_path: Vec<usize>,
+        block_number: u64,
+    ) -> QueryResult<()> {
+        let pos = proof.index().clone();
+        // First node in the merkle path is always the leaf node for which membership proof is provided
```
jbearer, line 1211: I think this could be `MerkleNode::Empty` in the case that an element was deleted from the tree by a state update (e.g. for the fee state, we delete an account when its balance reaches 0)

src/data_source/storage/sql.rs
```diff
@@ -1170,6 +1192,401 @@ where
...
+        let mut leaf_commit = Vec::new();
+        // Serialize the leaf node hash value into a vector
+        value
+            .serialize_compressed(&mut leaf_commit)
+            .map_err(ParseError::Serialize)?;
+
+        // Define the LTREE path to be inserted for the leaf node.
+        // Note:
+        // The traversal path is from the leaf to the root.
+        // The LTREE path is created by appending 'R' for the Root node to this path and then reversing it.
```
jbearer, line 1227: Why do we need to represent the root explicitly? It's a common ancestor of every path so could just be represented by the empty ltree.

src/data_source/storage/sql.rs
```diff
@@ -1170,6 +1192,401 @@ where
...
+                    // BitVec is used to separate out Empty children positions
+                    let mut children_bitvec = BitVec::new();
+                    for child in children {
+                        let child = child.as_ref();
+                        match child {
+                            MerkleNode::Empty => {
+                                children_bitvec.push(false);
+                            }
+                            MerkleNode::Branch { .. } => (),
+                            MerkleNode::Leaf { .. } => (),
```
jbearer, line 1296: It seems like we would need to push a `true` bit in these cases, but I think the reason why it's ok is because these cases are never reached, since in a merkle path the children are always forgotten. Still, I think we should either handle these cases properly or return an error, not silently ignore them

src/data_source/storage/sql.rs
```diff
@@ -1170,6 +1192,401 @@ where
...
+                sql_param(&serde_json::to_value(&pos).map_err(ParseError::Serde)?),
+                sql_param(&serde_json::to_value(elem).map_err(ParseError::Serde)?),
+            ]],
+        )
+        .await?;
+
+        let mut traversal_path = traversal_path.into_iter();
+
+        // Insert the internal nodes using the membership proof.
+        for node in proof.merkle_path().iter() {
```
jbearer, line 1265: I wonder if it is possible/worthwhile to collect the values generated in this loop and do a bulk `INSERT` at the end. I think as is we could easily end up doing like 20 or 30 separate statements per update.

src/data_source/storage/sql.rs
```diff
@@ -1170,6 +1192,401 @@ where
...
+            .collect::<Vec<_>>();
+
+        let created = match snapshot {
+            Snapshot::<Types>::Commit(commit) => {
+                let commit: &[u8; 32] = commit.as_ref();
+                // Get the block height using the merkle commitment.
+                let query = self
+                    .query_one(
+                        &format!(
+                            "SELECT n.created FROM {state_type} n INNER JOIN Header h on
```
jbearer, line 1380: I don't think a `JOIN` is necessary here, since `n.created` is the same as `h.height`

src/data_source/storage/sql.rs
```diff
@@ -1170,6 +1192,401 @@ where
...
+                            .map_err(ParseError::Deserialize)?,
+                        pos: serde_json::from_value(index.clone()).map_err(ParseError::Serde)?,
+                        elem: serde_json::from_value(entry.clone()).map_err(ParseError::Serde)?,
+                    });
+                }
+            }
+        }
+
+        Ok(proof_path)
+    }
```
jbearer, line 1503: We're missing the check that data isn't missing: https://www.notion.so/espressosys/Efficient-storage-and-queries-for-Merklized-state-8a8ccd79b73d4a7abe9a8df4208983dd?pvs=4#6a1b66321c47486bb318bd14d78afc29

src/data_source/storage/sql.rs
```diff
@@ -1170,6 +1192,401 @@ where
...
+                created
+            }
+        };
+
+        //  Get all the nodes in the path to the index.
+        // Order by pos DESC is to return nodes from the leaf to the root
+        let nodes = self
+            .query(
+                &format!(
+                    " WITH recent_nodes AS (
```
jbearer, line 1408: What's the reason for not using `SELECT DISTINCT ON` to get the most recent version of each node in the path? IMO that is easier to read than this subquery/self-join

src/data_source/storage/fs.rs
```diff
@@ -571,3 +600,11 @@ where
         Ok(res)
     }
 }
+
+#[async_trait]
+impl<Types: NodeType> MerklizedStateDataSource<Types> for FileSystemStorage<Types>
```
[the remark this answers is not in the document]
imabdulbasit: I think I implemented it because of Mock data

src/data_source/storage/sql.rs
imabdulbasit: I think distinct on just selects a random node and the results are then results are ordered by which is wrong.

src/merklized_state/data_source.rs
imabdulbasit: yeah this sounds way better

src/data_source/storage/sql.rs
jbearer: I think distinct on respects the order by clause. From [the docs](https://www.postgresql.org/docs/current/sql-select.html#SQL-DISTINCT):

> Note that the “first row” of each set is unpredictable **unless ORDER BY is used to ensure that the desired row appears first**

They given an example of doing pretty much exactly what we're trying to do

imabdulbasit: aah okay, one of the test was failing because of query but I think that was related to something else. I'll change it back

src/merklized_state/data_source.rs
imabdulbasit: so we would need to remove the state update from the update function currently used for updating the availability data

src/data_source/extension.rs
```diff
@@ -276,6 +279,62 @@ where
...
+#[async_trait]
+impl<D, U, Types, State> MerklizedStateDataSource<Types, State> for ExtensibleDataSource<D, U>
+where
+    D: MerklizedStateDataSource<Types, State> + Send + Sync,
+    U: Send + Sync,
+    Types: NodeType,
+    State: MerklizedState<Types> + 'static,
+    State::Index: Send + Sync,
+    State::Element: Send + Sync,
+    State::NodeValue: Send,
```
jbearer, line 291: Might be cleaner to put these trait bounds on the definitions of the associated types so we don't have to repeat them at every usage

src/merklized_state/data_source.rs
```diff
@@ -0,0 +1,126 @@
...
+pub trait MerklizedStateDataSource<Types, State>
+where
+    Types: NodeType,
+    State: MerklizedState<Types>,
+{
+    async fn get_path(
+        &self,
+        _state_type: &'static str,
+        _tree_height: usize,
+        _header_state_commitment_field: &'static str,
```
jbearer, line 46: `state_type` and `_header_state_commitment_field` should come from `State` (either associated constants or associated functions of the `MerklizedState` trait). It shouldn't be up to the caller to provide them since they are required to be consistent with `State` anyways.

What is `tree_height` needed for?

Also, you can remove the leading `_`s, it shouldn't warn since there is no default impl

src/merklized_state/data_source.rs
```diff
@@ -0,0 +1,126 @@
...
+        _snapshot: Snapshot<State>,
+        _key: String,
```
jbearer, line 48: `key` here should be the strongly typed `State::Key`. The caller (ie API endpoint handler) should be responsible for parsing from a string.

src/merklized_state/data_source.rs
```diff
@@ -0,0 +1,126 @@
...
+    ) -> QueryResult<MerklePath<State::Element, State::Index, State::NodeValue>>;
+}
+
+// This trait defines methods for updating the storage with the merkle tree state.
+#[async_trait]
+pub trait UpdateStateData<Types: NodeType, State: MerklizedState<Types>>: Send + Sync
+where
+    State::Element: Serialize,
+    State::Index: Serialize,
```
jbearer, line 57: These probably make sense as bounds on the associated type definition so we don't have to put them in every implementation

src/merklized_state/data_source.rs
```diff
@@ -0,0 +1,126 @@
...
+{
+    async fn insert_merkle_nodes(
+        &mut self,
+        _name: &'static str,
```
jbearer, line 61: Get this from `State`

src/merklized_state/data_source.rs
```diff
@@ -0,0 +1,126 @@
...
+use serde::Serialize;
+use std::fmt::Debug;
+
+use typenum::Unsigned;
+
+use std::{cmp::Ordering, sync::Arc};
+
+use crate::QueryResult;
+
+// This trait defines methods that a data source should implement
```
jbearer, line 34: Use doc comments `///`

src/data_source/fetching.rs
```diff
@@ -539,6 +545,39 @@ where
     }
 }
 
+#[async_trait]
```
jbearer, line 548: I think we need a similar blanket `UpdateStateData` here, like we have in `ExtensibleDataSource`. Basically anything that wraps an inner data source should blanket implement these traits, just in case the inner data source happens to implement them

src/merklized_state.rs
```diff
@@ -0,0 +1,137 @@
...
+    #[arg(
+        long = "merklized-state-extension",
+        env = "HOTSHOT_MERKLIZED_STATE_EXTENSIONS",
+        value_delimiter = ','
+    )]
+    pub extensions: Vec<toml::Value>,
+}
+
+#[derive(Clone, Debug, From, Snafu, Deserialize, Serialize)]
+pub enum Error {
```
jbearer, line 57: We should also have a variant wrapping `QueryError`, whose `status()` is `source.status()`. That allows us to return 404 e.g. if a query fails because some data was missing, rather than 500

src/data_source/storage/sql.rs
```diff
@@ -1170,6 +1188,491 @@ where
     }
 }
 
+#[async_trait]
+impl<Types: NodeType, State: MerklizedState<Types>> UpdateStateData<Types, State> for SqlStorage
+where
+    State::Index: Serialize + Send + Sync + Display + FromStr + 'static,
+    State::Element: Serialize + Send + Sync + 'static,
+    State::NodeValue: Send + 'static,
```
jbearer, line 1196: Should be able to remove this by adding trait bounds where the associated types are defined

src/data_source/storage/sql.rs
```diff
@@ -1170,6 +1188,491 @@ where
...
+                    let mut leaf_commit = Vec::new();
+                    // Serialize the leaf node hash value into a vector
+                    value
+                        .serialize_compressed(&mut leaf_commit)
+                        .map_err(ParseError::Serialize)?;
+
+                    // Define the LTREE path to be inserted for the leaf node.
+                    // Note:
+                    // The traversal path is from the leaf to the root.
+                    // The LTREE path is created by reversing the traversal path. Root node is represented as an empty LTree
```
jbearer, line 1240: I think this comment would make more sense at the definition of `LTree::from` where this logic is actually implemented

src/merklized_state/data_source.rs
jbearer: Oh, I see, `tree_height` is an argument of `to_traversal_path`. That's annoying, but oh well.

src/data_source/storage/sql.rs
```diff
@@ -1170,6 +1188,491 @@ where
...
+                commit
+                    .serialize_compressed(&mut writer)
+                    .map_err(ParseError::Deserialize)?;
+
+                // Get the block height using the merkle commitment.
+                let query = self
+                    .query_one(
+                        &format!(
+                            "SELECT created FROM Header where 
+                             h.data->>{header_state_commitment_field}"
```
jbearer, line 1407: Should be `height` instead of `created` I think, and looks like we're missing an `= $1` or something?

src/data_source/storage/sql.rs
```diff
@@ -1170,6 +1188,491 @@ where
...
+                        })
+                        .collect::<QueryResult<Vec<_>>>()?;
+                    // Use the Children merkle nodes to reconstruct the branch node
+                    proof_path.push(MerkleNode::Branch {
+                        value: State::NodeValue::deserialize_compressed(value.as_slice())
+                            .map_err(ParseError::Deserialize)?,
+                        children: child_nodes,
+                    });
+                } else {
+                    // If the children for a node does not exist then the node is actually a leaf node
```
jbearer, line 1524: What if it's `Empty`?

src/data_source/storage/sql.rs
```diff
@@ -1170,6 +1188,491 @@ where
...
+        // The merkle tree commitment from the header for the snapshot should match the root node hash in the table
+        // The last node in nodes vector is the root node
+        let root_node = nodes.last().ok_or(QueryError::Error {
+            message: "No node found in the db".to_owned(),
+        })?;
+        // Get the hash value
+        let root_hash_value = hashes.get(&root_node.hash_id).ok_or(QueryError::Error {
+            message: "Hash value for the root node not found".to_owned(),
+        })?;
+        if *root_hash_value != merkle_commitment {
```
jbearer, line 1478: I think there's something kind of annoying here. The `merkle_commitment` we store in the header is not just the root node hash, it contains some other metadata needed to fully commit to the structure of the tree, like the number of leaves. It's actually a serialization of [this struct](https://github.com/EspressoSystems/jellyfish/blob/main/primitives/src/merkle_tree/internal.rs#L101). So to pull out the root digest for this comparison, we need to parse `merkle_commitment` (it should be a tagged base 64 string) as a `MerkleTreeCommitment` using `serde_json::from_str`. Then you can check like `*root_hash_value == comm.digest()`

src/data_source/storage/sql.rs
```diff
@@ -1170,6 +1188,518 @@ where
...
+                                message: format!("failed to deserialize hash to Nodevalue err {e}"),
+                            }
+                        })
+                    } else {
+                        Ok(State::T::default())
+                    }
+                })
+                .collect::<QueryResult<Vec<_>>>()?;
+
+            State::Digest::digest(&root_children_values).map_err(|e| QueryError::Error {
```
jbearer, line 1481: Ayy this is annoying. I think this check is still not quite right (might give false positives): because we are recomputing the root hash based on the hashes of children saved in the database, the check will pass as long as we have the correct version of the immediate children of the root, even if nodes farther down the path are missing their latest version.

I think what we actually need to check is that the entire path is a valid merkle path and the value of the root node is consistent with the header field. So basically traverse each node in the path, recompute it's digest based on its children's digests and compare it to the value field of the node. Then finally check the root against `merkle_commitment`.

src/data_source/storage/sql.rs
```diff
@@ -2211,4 +2821,314 @@ mod test {
...
+
+            let (_, proof) = test_tree.lookup(i).expect_ok().unwrap();
+
+            tracing::info!("merkle path {:?}", merkle_path);
+
+            // merkle path from the storage should match the path from test tree
+            assert_eq!(merkle_path, proof.proof.clone(), "merkle paths mismatch");
+        }
+
+        // Lets insert some index again with different block height e.g 0
```
jbearer, line 2900: Comment appears wrong since we are actually inserting with block height 2

src/data_source/storage/sql.rs
```diff
@@ -2211,4 +2821,314 @@ mod test {
...
+        // There should be only 2 versions of this node
+        assert!(nodes.len() == 2, "incorrect number of nodes");
+        assert_eq!(nodes[0].created, 1, "wrong block height");
+        assert_eq!(nodes[1].created, 2, "wrong block height");
+
+        // Now we can have two snapshots for Index 0
+        // One with created = 1 and other with 2
+        // Query snapshot with created = 2
+
+        let path_2 = <SqlStorage as MerklizedStateDataSource<_, MockMerkleTree>>::get_path(
```
jbearer, line 2959: Need to `get_path` with `Index(1)` to check that the query for the old version still works

src/data_source/storage/sql.rs
```diff
@@ -2211,4 +2821,314 @@ mod test {
...
+                &storage,
+                &test_tree,
+                Snapshot::Index(2_u64),
+                0,
+            )
+            .await
+            .unwrap();
+        // Assert that the paths from the db and the tree are equal
+        assert_eq!(non_membership_path, proof.proof, "paths dont match");
+    }
```
jbearer, line 3075: Lookup up again with created = 1 and ensure we can still retrieve the membership proof from that block height

src/data_source/storage/sql.rs
jbearer: I think we are also missing a test for the case where this check fails

src/data_source/storage/sql.rs
```diff
@@ -1560,6 +1520,58 @@ where
...
+                        if data[*branch as usize] != val {
+                            return Err(QueryError::Error {
+                                message: "Some earlier state was used to calculate root commitment"
+                                    .to_string(),
+                            });
+                        }
+
+                        State::Digest::digest(&data).map_err(|_| QueryError::Missing)
+                    }
+                    MerkleNode::Empty => Ok(init),
```
jbearer, line 1560: Isn't it an error if we encounter an empty node in the middle of the traversal path? I thought it should be branch nodes only for the internal nodes.

src/merklized_state/data_source.rs
```diff
@@ -49,7 +49,7 @@ where
 {
     async fn get_path(
         &self,
-        state: &State,
+        tree_height: usize,
```
jbearer, line 52: Nit: I think this falls in the same category as the table and column names, where it is an implicit property of data stored in the database, and thus it must be fixed for any given instantiation of this API (or it requires a restart/migration to update). Thus it could probably also be a static method of `MerklizedState

src/data_source/storage/sql.rs
imabdulbasit: There was a case where one index was removed and the non membership proof had all empty nodes

src/merklized_state/data_source.rs
imabdulbasit: MerkleTreeScheme trait has a non static method for height so I was avoiding another method for height but it makes it cleaner so I'll add it

src/data_source/storage/sql.rs
imabdulbasit: Non membership can have empty nodes? I did remove one index from the tree and the proof had empty nodes

jbearer: Gotcha, this seems fine then

jbearer (approved):

PR #474, opened 2024-03-14
update hotshot tag and point to hotshot types

move47:
It also accommodates changes recently introuduced to Hotshot, especially `HotshotConfig`. Moreover, It makes `hotshot-types` points to its own repo, instead of hotshot crate.


src/node.rs
```diff
@@ -245,7 +245,10 @@ mod test {
             let EventType::Decide { leaf_chain, .. } = event.event else {
                 continue;
             };
-            for (leaf, vid) in leaf_chain.iter().rev() {
```
tbro, line 248: Seem to be a lot of cloning going on. I think you could de-structure here and remove some of those. 

```
for LeafInfo { leaf, vid, .. } in leaf_chain.iter().rev()
```
Looks like things are `Arc`ed so maybe it doesn't matter?

move47: Done. Yes, I thought the same, but for visibility I think this makes more sense.

tbro (approved):

PR #475, opened 2024-03-14
Reduce number of files generated by file system storage

jbearer:
* Update atomic store and enaable pruning of old archive files
* Increase file fill size for `LedgerLog`


nyospe (approved):
LGTM.

PR #507, opened 2024-04-02
Fix: merkle tree commitment parsing

imabdulbasit:
Header data is serialized as serde_json::Value when inserting it into the header table


nomaxg (approved):
Nice - can we add a regression test somewhere?

imabdulbasit:
> Nice - can we add a regression test somewhere?

This is only needed because we parse header as serde_json::Value. Going to add a test in sequencer

PR #512, opened 2024-04-04
Add explorer API for use with Block Explorers

Ayiga:
In order for the Block Explorer to function to the fullest extent of its design, specific functionality is needed.  Some initial design requirements are as follows:

- Load metric data for displaying latest block metrics
image

- Display Statistics overview of the entire block chain:
image

- Perform predictive auto-complete / search:
image

- Retrieve Page results in a single request with minimal page load time

To achieve that end it would be beneficial to create an API that caters to these needs directly.  This API would need to have full and quick access to historical data.  It makes sense for it to be died to an archival data source with fast relational lookup.  This Explorer API is only implemented for an SQL DataSource that is also an Availability Data Source.

The changes made for this pull request fulfill this contract in some initial terms that can be adjusted as needed later down the line.  It aims to target the API design as outlined in this [https://www.notion.so/espressosys/Hot-Shot-Query-Service-Block-Explorer-Statistics-a41308b1cc874661946bb597e051606b#719231f225674fc28536211aca5059b2](Notion Document, describing the API).

This can change in the future, but for the moment the primary target of this implementation is what will be targeted with the Cappuccino release.


src/data_source/storage/sql.rs
```diff
@@ -1787,6 +1789,636 @@ impl SqlStorage {
...
+        &self,
+    ) -> QueryResult<explorer::data_source::ExplorerSummary<Types>> {
+        let histograms= async {
+            let historgram_query_result = self.query(
+                "SELECT
+                    h.height AS height,
+                    h.timestamp AS timestamp,
+                    h.timestamp - lead(timestamp) OVER (ORDER BY h.height DESC) AS time,
+                    p.size AS size,
+                    (SELECT COUNT(*) AS transactions FROM transaction AS t WHERE t.block_height = h.height) as transactions
```
jbearer, line 2226: Should we just add `num_txns` to the payload table?

src/data_source/storage/sql.rs
```diff
@@ -1787,6 +1789,636 @@ impl SqlStorage {
...
+                ).await;
+
+            histograms
+        }.await?;
+
+        let genesis_overview = async {
+            let row = self
+                .query_one(
+                    "SELECT
+                    (SELECT COUNT(*) FROM header) AS blocks,
```
jbearer, line 2304: Better way to get number of blocks since genesis is `SELECT max(height) FROM header`: works even if some blocks are missing and is much faster

Ayiga: This is a very fair!

src/data_source/storage/sql.rs
Ayiga: We *could*, it would prevent us from having to join, but we will likely not avoid the joins in the future when attempting to pull number of transactions submitted by each rollup.  So while it saves us here, it doesn't prevent all future cases, I believe.

What do you think?

src/data_source/storage/sql.rs
```diff
@@ -1787,6 +1789,636 @@ impl SqlStorage {
...
+            explorer::data_source::TransactionIdentifier::Latest => (
+                format!(
+                    "SELECT {BLOCK_COLUMNS}
+                            FROM header AS h
+                            JOIN payload AS p ON h.height = p.height
+                            WHERE h.height = (
+                                SELECT t1.block_height
+                                    FROM transaction AS t1
+                                    ORDER BY (t1.block_height, t1.index) DESC
+                                    LIMIT 1
```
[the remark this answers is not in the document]
Ayiga: Agreed.

src/data_source/storage/sql.rs
```diff
@@ -1787,6 +1789,636 @@ impl SqlStorage {
...
+                        "SELECT {BLOCK_COLUMNS}
+                           FROM header AS h
+                           JOIN payload AS p ON h.height = p.height
+                           WHERE h.height IN (
+                                SELECT t1.block_height
+                                    FROM transaction AS t1
+                                    WHERE t1.block_height = $2
+                                    ORDER BY (t1.block_height, t1.index) DESC
+                                    LIMIT $1
+                            )
```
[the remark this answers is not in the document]
Ayiga: This is true, and will change this.

The limit is meant to ensure that we only select the `block`s that we need, but since the `block` has a further constraint of just that single `block` itself the only returned value can possibly be just that single `block`.

src/data_source/storage/sql.rs
```diff
@@ -1787,6 +1789,636 @@ impl SqlStorage {
...
+
+        let result = self.query(&query, params).await.map(|row_stream| {
+            row_stream.map(|row| -> Result<BlockSummary<Types>, QueryError> {
+                let block = parse_block::<Types>(row?)?;
+                Ok(BlockSummary::try_from(block)?)
+            })
+        });
+
+        let results =
+            result.map(|res| async move { res.try_collect::<Vec<BlockSummary<Types>>>().await });
```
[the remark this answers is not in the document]
Ayiga: Yes! I missed the branch operator on this!

src/data_source/storage/sql.rs
```diff
@@ -1787,6 +1789,636 @@ impl SqlStorage {
...
+                    Box::new(height as i64),
+                    Box::new(request.num_blocks.get() as i64),
+                ],
+            ),
+            explorer::data_source::BlockIdentifier::Hash(hash) => (
+                format!(
+                    "SELECT {BLOCK_COLUMNS}
+                        FROM header AS h
+                        JOIN payload AS p ON h.height = p.height
+                        WHERE h.hash <= (SELECT h1.height FROM header AS h1 WHERE h1.hash = $1)
```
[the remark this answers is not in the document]
Ayiga: I believe your proposed change was meant to the intention.  Good call!

src/data_source/fetching.rs
```diff
@@ -682,6 +684,126 @@ where
     }
 }
 
+#[async_trait]
+impl<Types, S, P> ExplorerDataSource<Types> for FetchingDataSource<Types, S, P>
+where
+    Types: NodeType,
+    Payload<Types>: QueryablePayload,
+    Header<Types>: QueryableHeader<Types> + explorer::traits::ExplorerHeader<Types>,
+    S: ExplorerStorage<Types> + Send + Sync,
```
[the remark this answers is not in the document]
Ayiga: I suppose I misunderstood the purpose for the `DataSource` versus `Storage` distinction.  I figured it was primarily meant as a means of separating the two concepts, and didn't have a deeper intention beyond that.  I can absolutely get rid of the `ExplorerDataSource` in favor of the `ExplorerStorage`.

src/data_source/storage/sql.rs
```diff
@@ -1787,6 +1789,636 @@ impl SqlStorage {
...
+    async fn get_transaction_summaries(
+        &self,
+        request: &explorer::data_source::GetTransactionSummariesRequest<Types>,
+    ) -> QueryResult<Vec<explorer::data_source::TransactionSummary<Types>>> {
+        let range = &request.range;
+        let target = &range.target;
+        let filter = &request.filter;
+
+        let (query, params, offset): (String, Vec<Box<dyn ToSql + Send + Sync>>, Option<usize>) =
+            match (target, filter) {
```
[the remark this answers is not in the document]
Ayiga: A block stream could be a great place to start in most of these query places as they are essentially streaming from a given block.  In the majority of cases we are able to determine the starting block to a simple (block, transaction offset) combination, and then we just proceed backwards from there.  There is a case where this doesn't hold, and that's for Rollup filtering (which isn't implemented yet), but that can be handled separately if it helps to simplify the overall logic.

I'm not so certain we'd want to benefit from the existing stream implementations, as this is using the `FetchingDataSource` which will load any missing `block`s that are encountered, whereas the `Explorer` isn't attempting to have this side-effect.  It wants to just report things as they are rather than trying to proactively fetch missing data.

The hand-crafted queries in each case are primarily just a means to avoid any potential for SQL injections by eliminating any possibility of concatenation.  It also potential provides a few benefits, the first being that if we wish to pursue a prepared statement approach, these static queries could be pre-submitted to the SQL service at launch, and validated before the rest of the program engages.  This would potentially have an added benefit of slightly faster query processing.  In addition, static queries could be prefixed with an `SQL comment` that could aid in identification of the query using `SQL statement` logs.

However, the flow of the query in general is figuring out what `block`s to target, and then returning them.  This could potentially be split up into separate queries, and while we'd suffer speed due to performing multiple queries we may have gains in readability and maintainability as a result, and this seems like a fair price to pay.

src/data_source/storage/sql.rs
```diff
@@ -1787,6 +1789,636 @@ impl SqlStorage {
...
+    explorer::data_source::BalanceAmount<Types>: Into<explorer::monetary_value::MonetaryValue>,
+{
+    async fn get_block_summaries(
+        &self,
+        request: &explorer::data_source::GetBlockSummariesRequest<Types>,
+    ) -> QueryResult<Vec<explorer::data_source::BlockSummary<Types>>> {
+        let request = &request.0;
+
+        let (query, params): (String, Vec<Box<dyn ToSql + Send + Sync>>) = match request.target {
+            explorer::data_source::BlockIdentifier::Latest => (
```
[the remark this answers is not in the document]
Ayiga: This is a fair point.  I was aiming for reducing the potential for naming collisions by being explicit with the specific types here.  I think there may be future potential for these names to be duplicated between sub-crates / modules of the `HotShot Query Service`.  But this is not necessary at the moment.

src/data_source/storage/sql.rs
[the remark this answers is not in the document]
Ayiga: I'm actually having trouble pursuing the block stream approach as we run into some pitfalls that the current separate queries are resolving on their own.

The first is that we only pull the `block`s that have corresponding `transaction`s which prevents us from neeing to check and expand every block to see if we have the required number of transactions available.

The second is that we only pull the number of blocks we need to actually pull `N` transactions, which is what a lot of the sub-queries are doing in these cases.

As not every query takes the same number of sub-query paths it's difficult to resolve the nuances of everything with the same manner.  Usually the inner most query is utilized to determine the starting point (where applicable).  The second sub query then is used to limit the number of transactions that will be pulled up to the required `limit`.  This second sub-query can incur an addition filter for the `TransactionSummaryFilter` which helps to reduce the `block`s returned even further.  After that the `block`s are returned from only those transactions that this applies to.

Since the other variants of the queries are able to be simplified, this can make it quite difficult to reason about these further reductions that would apply.  This may require some further consideration.

src/data_source/storage/sql.rs
```diff
@@ -1789,6 +1800,534 @@ impl SqlStorage {
...
+    ) -> QueryResult<Vec<TransactionSummary<Types>>> {
+        let range = &request.range;
+        let target = &range.target;
+        let filter = &request.filter;
+
+        // We need to figure out the transaction target we are going to start
+        // returned results based on.
+        let transaction_target = match target {
+            TransactionIdentifier::Latest => self.query_one::<str, Vec<Box<dyn ToSql + Send + Sync>>>(
+                "SELECT t.block_height AS height, t.index AS index FROM transaction AS t ORDER BY (t.block_height, t.index) DESC LIMIT 1",
```
imabdulbasit, line 1931:
```suggestion
                "SELECT t.block_height AS height, t.index AS index FROM transaction AS t ORDER BY (t.block_height, t.index) DESC LIMIT 1",
```

src/data_source/storage/sql.rs
```diff
@@ -1789,6 +1800,534 @@ impl SqlStorage {
...
+                vec![],
+            ).await,
+            TransactionIdentifier::HeightAndOffset(height, offset) => self.query_one(
+                "SELECT t.block_height AS height, t.index AS index FROM transaction AS t WHERE t.block_height = $1 ORDER BY (t.block_height, t.index) DESC OFFSET $2 LIMIT 1",
```
imabdulbasit, line 1935: I think block_height column in transactions should have a separate  index

src/data_source/storage/sql.rs
```diff
@@ -1789,6 +1800,534 @@ impl SqlStorage {
...
+                            ORDER BY h.height DESC"
+                ),
+                vec![],
+            ),
+            TransactionIdentifier::HeightAndOffset(height, offset) => (
+                format!(
+                    "SELECT {BLOCK_COLUMNS}
+                            FROM header AS h
+                            JOIN payload AS p ON h.height = p.height
+                            WHERE h.height = (
```
imabdulbasit, line 2083: the subquery can return multiple rows so this should use **IN** clause?

src/data_source/storage/sql.rs
```diff
@@ -1789,6 +1800,534 @@ impl SqlStorage {
...
+            histograms,
+        })
+    }
+
+    async fn get_search_results(&self, query: String) -> QueryResult<SearchResult<Types>> {
+        let block_query = format!(
+            "SELECT {BLOCK_COLUMNS}
+                FROM header AS h
+                JOIN payload AS p ON h.height = p.height
+                WHERE h.hash LIKE $1 || '%'
```
imabdulbasit, line 2279: this would be an expensive query. Maybe we can use gist/gin index for this ? https://www.postgresql.org/docs/current/pgtrgm.html @jbearer

src/data_source/storage/sql.rs
jbearer: I think it's a nice and reasonable thing to have in the db in any case. But not a super high priority, it doesn't need to block this PR

src/data_source/storage/sql.rs
jbearer: Hm I thought btree index was applicable in the special case where you have `LIKE 'something%' (because that corresponds to lexicographic ordering).

imabdulbasit: This query is for full text search or just searching the prefix?

Ayiga: It should just be a prefix search I believe.

Usually when doing FTS you end up with a keyword like `MATCH` or something.  Though, I'm not as certain if this is true in `postgres`.

Ayiga: Yeah, I was able to confirm that `Full Text Search` has its own functions and calling symantics, in addition to requiring specific index setup.  Apparently there are multiple versions, the one I read about specifically was `to_tsquery` for it's `MATCH` command.

the `||` in this case is the concatenation operator, though the `CONCAT` function exists, I believe in `postgres`, perhaps changing that could help prevent any confusion?

src/data_source/storage/sql.rs
Ayiga: It currently does not look to have a separate one.  It only has an index on the `hash`, and on the combination of `(block_height, index)`.

Unless I'm misunderstanding this comment and you're suggesting that we should add a new index for just `block_height`.

imabdulbasit: yes we should add a separate btree index for blockheight

src/data_source/storage/sql.rs
Ayiga: There's a `LIMIT 1` on this subquery, so this subquery can only return a single row.

src/data_source/storage/sql.rs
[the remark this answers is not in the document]
Ayiga: I was able to get this working such that I'm not matching on both `TransactionIdentifier` and `TransactionSummaryFilter` at the same time.   If we consider `TransactionIdentifier` has having `N` enum variants and `TransactionSummaryFilter` has having `M` variants, the previous approach would have required `N * M` hard-coded queries.

With the modified approach, I instead run a separate query for each, resulting in the need for `N + M` unique queries instead.

src/data_source/storage/sql.rs
Ayiga: In my local testing of the database for the sequencer, adding an index for this yields no speed up, and in fact seems to make things marginally slower.  I figure since there is already a compound index for `(block_height, index)` that the `block_height` is kind of already indexed in this situation.  Though my experience with using partially applied compound indexes tends to indicate that this isn't usually the case.

Sampling the performance by running this query:
```sql
SELECT t.block_height AS height, t.index AS index FROM hotshot.transaction AS t WHERE t.block_height = 20 ORDER BY (t.block_height, t.index) DESC OFFSET 0 LIMIT 1
```

If I run the query early (with a minimal amount of blocks saved to the database), I do get a more complex query plan:
image

But this is likely fine, as there really isn't enough data to really speed up any execution via the query that a linear scan of fewer than 1500 indexes would still be fairly fast, taking at worst `83ms` on my hardware.

Once the demo stores above 1500 indexes the query plan changes:
image

And looks fairly similar to the query plan from the added index.  It takes around `36.1ms` on average.

If add the index via:
```sql
CREATE INDEX transaction_block_height ON hotshot.transaction (block_height)
```

The query plan does in fact change:
image

Which takes about `43.8ms` on average to execute with the index added.

So we get slower queries, but for a "cheaper" query cost.

I'm not sure what to make of that.  What do you think @imabdulbasit?

imabdulbasit: I think it is using the index because the first column in compound index is block height. I have had better read performance when adding separate index to table having millions of rows but that also slows down the upserts. I am keeping a note of this and going to investigate when we have a large db. Resolving this for now.

src/data_source/fetching.rs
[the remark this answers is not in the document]
Ayiga: `ExplorerDataSource` removed in commit `1e7a0ea7ba1222fd16c24922872c0c9a16a8ca51`

src/data_source/storage/sql.rs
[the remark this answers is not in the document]
Ayiga: This change is implemented in commit `4fd964307e8c6b25dc0c76ccb8a88f41f3973901`

src/data_source/storage/sql.rs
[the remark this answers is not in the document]
Ayiga: Adjusted in commit `f5486981d52d5826f9de5f4bd7bc30448e79abc0`

src/data_source/storage/sql.rs
Ayiga: This has been adjusted in commit `f5486981d52d5826f9de5f4bd7bc30448e79abc0`.

src/data_source/storage/sql.rs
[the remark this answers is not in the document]
Ayiga: This change was made in commit `1e7a0ea7ba1222fd16c24922872c0c9a16a8ca51`.

src/data_source/storage/sql.rs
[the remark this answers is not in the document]
Ayiga: This has been adjusted in commit `32d322f0b0e9afc6a85db5e1274caebbc800fd92`.

src/data_source/storage/sql.rs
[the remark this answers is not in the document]
Ayiga: This has been simplified in commit `f5486981d52d5826f9de5f4bd7bc30448e79abc0`.

src/data_source/storage/sql.rs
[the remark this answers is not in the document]
Ayiga: Since the queries on this have been reworked, this may no longer be applicable.

src/data_source/storage/sql.rs
```diff
@@ -1925,6 +1938,539 @@ impl SqlStorage {
...
+        });
+
+        let transaction_summary_vec = transaction_summary_stream
+            .try_collect::<Vec<TransactionSummary<Types>>>()
+            .await?;
+
+        Ok(transaction_summary_vec
+            .into_iter()
+            .skip_while(|txn| {
+                if let TransactionIdentifier::Hash(hash) = target {
```
imabdulbasit, line 2185: why do we need this check? Can we make target included as well in transaction summaries?

src/data_source/storage/sql.rs
```diff
@@ -1925,6 +1938,539 @@ impl SqlStorage {
...
+        let (offset, txn) = match target {
+            TransactionIdentifier::Latest => txns.into_iter().enumerate().last().unwrap(),
+            TransactionIdentifier::HeightAndOffset(_, offset) => {
+                txns.into_iter().enumerate().rev().nth(offset).unwrap()
+            }
+            TransactionIdentifier::Hash(hash) => txns
+                .into_iter()
+                .enumerate()
+                .find(|(_, txn)| txn.commit() == hash)
+                .unwrap(),
```
imabdulbasit, line 2265: we should handle errors instead of unwrapping

src/data_source/storage/sql.rs
Ayiga: I'm not entirely following what you're asking here about including the target in the `transaction summaries`.

I'm fairly certain we do need a check for the `Hash` here, as we retrieve `Block`s from the `query`, and we don't know where in the block that the `Transaction` identified by the `Hash` in question occurs.

src/data_source/storage/sql.rs
Ayiga: I think you might mean we should return an error instead of unwrapping?

imabdulbasit: yes

Ayiga: Adjusted in commit `17ad104c9da3002c2fe59ba908d462800dd59da6`

src/explorer/query_data.rs
```diff
@@ -0,0 +1,920 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+
+pub enum GetBlockDetailError {
+    #[serde(untagged)]
+    Unimplemented(Unimplemented),
+    #[serde(untagged)]
+    BlockNotFound(String),
+    #[serde(untagged)]
+    QueryError(QueryError),
+}
```
imabdulbasit, line 519:
```suggestion
  #[serde(untagged)]
pub enum GetBlockDetailError {
    Unimplemented(Unimplemented),
    BlockNotFound(String),
    QueryError(QueryError),
}
```

Ayiga: This has been adjusted in commit `c7dedab5219c8fa78f3cb7cac872cd2d2412ccb7`.

imabdulbasit (approved):
LGTM

PR #518, opened 2024-04-08
Update dependencies

rob-maron:
Includes HotShot, Jellyfish, commit, etc, and pulls in some of our deps via crates.io instead.


examples/simple-server.rs
```diff
@@ -151,7 +153,16 @@ async fn init_consensus(
         })
         .collect::<Vec<_>>();
 
+    // Pick a random port for the builder
+    let builder_port = pick_unused_port().expect("failed to get unused port");
+    let builder_url = surf_disco::Url::parse(&format!("127.0.0.1:{builder_port}"))
```
jbearer, line 158: Usually we use 0.0.0.0 for a binding address. I recall there was some OS-specific behavior with 127.0.0.1. Don't really remember what the specific issue was, though, so if this works I guess it's fine

src/fetching/provider/query_service.rs
```diff
@@ -67,7 +67,9 @@ where
         match res {
             Ok((payload, common)) => {
                 // Verify that the data we retrieved is consistent with the request we made.
-                let num_storage_nodes = VidSchemeType::get_num_storage_nodes(common.common());
+                let num_storage_nodes =
+                    usize::try_from(VidSchemeType::get_num_storage_nodes(common.common()))
```
jbearer, line 71: I think we can just do `as u32` here. IMO we will only run on 32 bit systems or larger and stuff like this just makes it harder to read thecode

jbearer (approved):
👍

PR #566, opened 2024-05-03
Automatically reset postgres connection when closed

### LINKED ISSUE #560: Refresh postgres connection if closed
If any of the low-level tokio-postgres methods fail due to a [closed connection](https://docs.rs/tokio-postgres/latest/tokio_postgres/error/struct.Error.html#method.is_closed), we should automatically replace the connection in the `SqlStorage` instance with a fresh one. The connection can be closed for a number of reasons, including the server restarting.

jbearer:
Closes #560

### DIFF
--- a/src/data_source/storage/sql/client.rs
+++ b/src/data_source/storage/sql/client.rs
@@ -1,0 +1,604 @@
+// Copyright (c) 2022 Espresso Systems (espressosys.com)
+// This file is part of the HotShot Query Service library.
+//
+// This program is free software: you can redistribute it and/or modify it under the terms of the GNU
+// General Public License as published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+// This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+// even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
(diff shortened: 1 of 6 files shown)


imabdulbasit:
Looks good! this just needs to be rebased to  include some new changes related to TmpDb

src/data_source/storage/sql/client.rs
```diff
@@ -0,0 +1,582 @@
...
+            .unwrap();
+        assert_eq!(modified, 2);
+        assert_eq!(get_values().await, [1, 2, 3, 4]);
+
+        // Shut down the database.
+        db.stop();
+
+        // Subsequent statements and queries within the transaction fail.
+        tracing::info!("query with DB shut down");
+        let err = client
```
imabdulbasit, line 474: if this transaction fails then shouldn't the `in_progress` be set to `false`. After this, if I begin a new transaction then that should create a new connection or return an error if that times out?

imabdulbasit: or maybe if I create a new transaction and the connection is closed then it retries ?

jbearer: Yes, but we haven't started a new transaction yet. We try that at the end of this test and make sure statements in the new transaction succeed

imabdulbasit (approved):
LGTM

PR #570, opened 2024-05-06
Optimize merkle path query

imabdulbasit:
Closes  https://github.com/EspressoSystems/hotshot-query-service/issues/569

### This PR:

Individual queries combined with a UNION operation are significantly faster and less expensive to execute compared to queries using the LTREE operation. 

- Removes LTREE 
- Batch insert children hashes and node hashes in one query

### Key places to review:

src/data_source/storage/sql.rs

### DIFF
--- a/src/data_source/storage/sql.rs
+++ b/src/data_source/storage/sql.rs
@@ -1584,55 +1584,6 @@ where
             proof: proof_path.into(),
         })
     }
-
-    async fn keys(&self, snapshot: Snapshot<Types, State, ARITY>) -> QueryResult<Vec<State::Key>> {
-        let state_type = State::state_type();
-
-        // Identify the snapshot.
-        let created = self.snapshot_info(snapshot).await?.0;
-
-        // Get all the nodes which correspond to an entry, ie have a non-NULL index field.
-        let rows = self
-            .query(
-                &format!(
-                    "SELECT DISTINCT ON (path) index
-                    FROM {state_type}
-                    WHERE created <= $1 AND index IS NOT NULL
-                    ORDER BY path, created DESC;"
-                ),
-                [sql_param(&created)],
-            )
-            .await?;
-        rows.map(|row| {
-            let row = row.map_err(|err| QueryError::Error {
-                message: format!("failed to fetch key: {err:#}"),
-            })?;
-            serde_json::from_value(row.get("index")).map_err(|err| QueryError::Error {
-                message: format!("failed to deserialize key: {err:#}"),
-            })
-        })
-        .try_collect()
-        .await
-    }
-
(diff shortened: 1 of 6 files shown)


src/data_source/storage/sql.rs
```diff
@@ -1598,10 +1595,10 @@ where
         let rows = self
             .query(
                 &format!(
-                    "SELECT DISTINCT ON (pos) index
+                    "SELECT DISTINCT ON (path) index
                     FROM {state_type}
                     WHERE created <= $1 AND index IS NOT NULL
-                    ORDER BY pos, created DESC;"
+                    ORDER BY path, created DESC;"
```
jbearer, line 1601: Does this `ORDER BY path` still do the right thing without using ltree? Tbh we've gotten rid of the need for this and `load_snapshot` in the sequencer now. We could delete these two functions and thus remove a lot of maintenance burden in this repo. If we ever want to add them back, we have the git history

src/data_source/storage/sql.rs
```diff
@@ -3259,16 +3236,16 @@ pub mod testing {
...
                 hash_id INT NOT NULL REFERENCES hash (id),
                 children INT[],
                 children_bitvec BIT(8),
                 index JSONB,
                 entry JSONB 
             );
-            ALTER TABLE {name} ADD CONSTRAINT {name}_pk PRIMARY KEY (pos, created);
-            CREATE INDEX {name}_path ON {name} USING GIST (pos);"
+            ALTER TABLE {name} ADD CONSTRAINT {name}_pk PRIMARY KEY (path, created);
+            CREATE INDEX {name}_created ON {name} (path);"
```
jbearer, line 3248: This seems like a typo, we have `_created` in the name but `(path)` in the index. Also, do we still need this index if the query now uses the primary key?

src/data_source/storage/sql.rs
jbearer: The idea of having a randomized test might be useful. We could keep that but just test `get_path` in it instead of `get_snapshot`

jbearer: I'll leave it up to you, if you want to keep this and it still works with the new way of doing it then I'm fine leaving it in

imabdulbasit: I think yes ORDER BY is working as expected for arrays too. For LTREE it is created as string 
image

jbearer (approved):

src/data_source/storage/sql.rs
imabdulbasit: addressed in `8d2acd8e17fe00b04015b8355bed91517605e1ad`

imabdulbasit: Yeah the query planner uses this index for some individual queries

PR #580, opened 2024-05-10
Remove `namespace_ids_for_offset` from `ExplorerHeader`

### LINKED ISSUE #579: `ExplorerHeader` should not have the method `namespace_ids_for_offset`
The method `namespace_ids_for_offset` is intended to get a list of `NamespaceId`s for a given Transaction offset on a Block Header.  Considering implementation details the Block Header doesn't actually have access to this information, based on the discussion here: [https://github.com/EspressoSystems/espresso-sequencer/pull/1428#discussion_r1596717521](https://github.com/EspressoSystems/espresso-sequencer/pull/1428#discussion_r1596717521).

Since this `method` is only utilized for `TransactionSummary`, `TransactionSummary` has access to the actual `Transaction` itself when being constructed, and the implementation details for a `Transaction` actually has access to its `NamespaceId`, then we can make this work by creating a `trait` for the `Transaction` itself to retrieve its own `NamespaceId`.

So, we need to remove `namespace_ids_for_offset`, and create a new `trait` for a `Transaction` for use with the `Explorer API`, that allows us to retrieve the `NamespaceId` from the transaction itself.

Ayiga:
Closes #579 

### This PR:

Removes the `namespace_ids_for_offset` method from the `ExplorerHeader` trait in favor of a `namespace_id` method on a new trait `ExplorerTransaction`.

### This PR does not:

Add any consideration for representing the `NamespaceId` on the `MockTransaction` struct.

### Key places to review:

src/explorer/traits.rs
This is where the new trait is defined, and the previous method is removed.

src/testing/mocks.rs
This is where the traits are implemented for testing

### DIFF
--- a/src/data_source/extension.rs
+++ b/src/data_source/extension.rs
@@ -17,7 +17,7 @@ use crate::{
         PayloadQueryData, QueryableHeader, QueryablePayload, TransactionHash, TransactionIndex,
         UpdateAvailabilityData, VidCommonQueryData,
     },
-    explorer::{self, ExplorerHeader},
+    explorer::{self, ExplorerHeader, ExplorerTransaction},
     merklized_state::{
         MerklizedState, MerklizedStateDataSource, MerklizedStateHeightPersistence, Snapshot,
         UpdateStateData,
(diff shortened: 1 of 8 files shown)


jbearer (approved):
LGTM

PR #583, opened 2024-05-12
Use computed columns for merkle tree roots

imabdulbasit:
Closes #<ISSUE_NUMBER>

### This PR:

uses computed columns for merkle tree roots

https://github.com/EspressoSystems/espresso-sequencer/pull/1451


migrations/V20__merkle_root_columns.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE header
```
jbearer, line 1: This migration should go in the sequencer

src/data_source/storage/sql.rs
```diff
@@ -3183,6 +3183,11 @@ pub mod testing {
                 value BYTEA  NOT NULL UNIQUE
             );
         
+
+            ALTER TABLE header
```
jbearer, line 3187: Could just include this in the `CREATE TABLE` statement

imabdulbasit: dont we need a separate migration for this?

imabdulbasit: this is for test merkle tree so didn't want to add it in migration

jbearer (approved):

PR #594, opened 2024-05-23
Fix explorer summary histogram data order

### LINKED ISSUE #593: Explorer histogram data is provided in the wrong order
The data provided for the Explorer UI Histograms is provided in descending block order, and it's meant to be in ascending block order in order to adhere to normal histogram display paradigms.

Ayiga:
Closes #593

### This PR:

The histogram data for the `get_explorer_summary` endpoint of the `Explorer API` is providing its data in the wrong order.  The data is meant to be delivered in `block height` ascending order.  Instead it's being delivered in `block height` descending order.  This pull request changes the sort order of the data being retrieved in the query itself, which corrects the issue.

### This PR does not:

Make any other significant changes to the logic.

### DIFF
--- a/src/data_source/storage/sql.rs
+++ b/src/data_source/storage/sql.rs
@@ -2219,8 +2219,10 @@ where
                 FROM header AS h
                 JOIN payload AS p ON
                     p.height = h.height
-                ORDER BY h.height DESC
-                LIMIT 50",
+                WHERE
+                    h.height IN (SELECT height FROM header ORDER BY height DESC LIMIT 50)
+                ORDER BY h.height ASC 
+                ",
                 Vec::<Box<dyn ToSql + Send + Sync>>::new(),
             ).await?;
 


src/data_source/storage/sql.rs
```diff
@@ -2219,8 +2219,10 @@ where
                 FROM header AS h
                 JOIN payload AS p ON
                     p.height = h.height
-                ORDER BY h.height DESC
-                LIMIT 50",
+                WHERE
+                    h.height IN (SELECT height FROM header ORDER BY height DESC LIMIT 50)
+                ORDER BY h.height ASC
```
jbearer, line 2224: I'm thinking it may be better to avoid the sub-query/`IN` predicate, instead just select by `height DESC LIMIT 50` and then reverse the order in memory. I'm not sure though.

src/data_source/storage/sql.rs
```diff
@@ -2219,8 +2219,10 @@ where
                 FROM header AS h
                 JOIN payload AS p ON
                     p.height = h.height
-                ORDER BY h.height DESC
-                LIMIT 50",
+                WHERE
+                    h.height IN (SELECT height FROM header ORDER BY height DESC LIMIT 50)
```
jbearer, line 2223: It occurs to me that this will give the latest 50 blocks we have, even if they're non-consecutive (ie if we are missing a header for some reason). What will happen to the histogram in this case? Will there be a blank slot/any indication that data is missing? If not it could be misleading

src/data_source/storage/sql.rs
Ayiga: I considered doing this... But for the sake of ease, I opted for the query since it only requires the change in a single place.  Though I am open to reversing it in memory as well.

src/data_source/storage/sql.rs
Ayiga: Indeed, no empty space would be inserted into the histogram, so it would show them as a contiguous list of blocks.  This may be potentially misleading, and could cause confusion upon inspection.

Ayiga: Though, it does beg the question of how would we want to represent missing data in a block histogram?

src/data_source/storage/sql.rs
jbearer: We can leave it unless/until it causes problems. It probably won't

src/data_source/storage/sql.rs
jbearer: We could punt on the UI issue by just failing the endpoint if data is missing (e.g. you can do a quick scan over the `height` column of each row returned from the database and return an error if not consecutive). The browser should retry and the endpoint should eventually succeed once the missing data has been fetched.

This issue was already present though, so it can be considered out of scope for this PR, we can just create an issue to address it later

jbearer (approved):

PR #599, opened 2024-05-29
add semaphore to ratelimit number of fetches

imabdulbasit:


src/fetching.rs
```diff
@@ -158,6 +163,8 @@ impl<T, C> Fetcher<T, C> {
             // Now we are responsible for fetching the object, reach out to the provider.
             let mut delay = min(MIN_RETRY_DELAY, max_retry_delay);
             let res = loop {
+                // Acquire a permit from the semaphore to rate limit the number of concurrent fetch requests
+                permit.acquire().await;
```
jbearer, line 167: If the fetch fails we need to drop the permit explicitly. Should not hold it while sleeping

src/fetching.rs
```diff
@@ -83,13 +85,15 @@ pub struct Fetcher<T, C> {
...
     retry_delay: Duration,
+    permit: Arc<Semaphore>,
 }
 
 impl<T, C> Default for Fetcher<T, C> {
     fn default() -> Self {
         Self {
             in_progress: Default::default(),
             retry_delay: DEFAULT_RETRY_DELAY,
+            permit: Arc::new(Semaphore::new(DEFAULT_RATE_LIMIT)),
```
jbearer, line 96: Add `with_rate_limit` and related configuration in `FetchingDataSource`

src/data_source/fetching.rs
```diff
@@ -900,6 +908,12 @@ where
             vid_common_fetcher = vid_common_fetcher.with_retry_delay(delay);
         }
 
+        if let Some(limit) = builder.rate_limit {
+            payload_fetcher = payload_fetcher.with_rate_limit(limit);
+            leaf_fetcher = leaf_fetcher.with_rate_limit(limit);
+            vid_common_fetcher = vid_common_fetcher.with_rate_limit(limit);
```
jbearer, line 914: I'm thinking instead of `limit` each fetcher should take an `Arc<Semaphore>`, so we can share the same semaphore between all of them. That way instead of a rate limit *per resource type* (which is kind of arbitrary) we just have an overall limit

src/data_source/fs.rs
```diff
@@ -158,10 +158,19 @@ where
...
     ///
     /// The [FileSystemDataSource] will manage its own persistence synchronization.
-    pub async fn create(path: &Path, provider: P) -> anyhow::Result<Self> {
-        FetchingDataSource::builder(FileSystemStorage::create(path).await?, provider)
-            .build()
-            .await
+    pub async fn create(
+        path: &Path,
+        provider: P,
+        fetch_rate_limit: Option<usize>,
```
jbearer, line 164: Remove this argument, if you want to configure a rate limit you should just create this from a builder

src/data_source/sql.rs
```diff
@@ -40,27 +40,32 @@ impl Config {
     pub async fn connect<Types, P: AvailabilityProvider<Types>>(
         self,
         provider: P,
+        fetch_rate_limit: Option<usize>,
```
jbearer, line 43: Same here

src/data_source/sql.rs
```diff
@@ -40,27 +40,32 @@ impl Config {
...
+            .build()
+            .await
     }
 
     /// Connect to the database, setting options on the underlying [`FetchingDataSource`] using the
     /// [`fetching::Builder`] interface.
     pub async fn builder<Types, P: AvailabilityProvider<Types>>(
         self,
         provider: P,
+        fetch_rate_limit: Option<usize>,
```
jbearer, line 61: And here

src/data_source/sql.rs
```diff
@@ -320,9 +325,16 @@ where
     /// underlying [`FetchingDataSource`], before constructing the [`SqlDataSource`] with
     /// [`build`](fetching::Builder::build). For a convenient constructor that uses the default
     /// fetching options, see [`Config::connect`].
-    pub async fn connect(config: Config, provider: P) -> Result<Builder<Types, P>, Error> {
-        let builder = Self::builder(SqlStorage::connect(config).await?, provider);
+    pub async fn connect(
+        config: Config,
+        provider: P,
+        fetch_rate_limit: Option<usize>,
```
jbearer, line 331: And here

src/data_source/fs.rs
imabdulbasit: we call the create function of FetchingDataSource in sequencer which creates a  default builder

jbearer: Then we should change that to use a builder

jbearer (approved):

PR #601, opened 2024-05-30
Update tide-disco

jbearer:


Cargo.toml
```diff
@@ -12,7 +12,7 @@
 
 [package]
 name = "hotshot-query-service"
-version = "0.1.23"
+version = "0.1.24"
```
imabdulbasit, line 15: This should be 0.1.27 as we have a tag for 0.1.26

imabdulbasit (changes requested):

Cargo.toml
jbearer: [36db5de](https://github.com/EspressoSystems/hotshot-query-service/pull/601/commits/36db5decbcbe1cfce78c720513d946b2535989ca)

imabdulbasit (approved):

PR #603, opened 2024-05-30
Rate limit the proactive fetching task

jbearer:
We now have rate limiting at the low-level fetcher, which is an excellent failsafe. However, by the time we get to that point, we have already consumed resources: memory for the list of waiters on the semaphore, memory for the task doing the fetch, memory for the fetch request itself, and memory for the list of callbacks on each item which is fetched for more than one purpose.

In most cases this is fine, because we just don't spawn that many active fetches at once. However, for proactive fetching, we may request thousands of active fetches at the same time, and even though we are now limiting the network traffic they generate, we are not limiting the amount of memory they consume.

This change adds a simple `await` instead of dropping/detaching the active fetch futures so that the proactive fetching task cannot proceed faster than the fetches can be completed. Note that we still get parallelism up to the chunk size.


rob-maron (approved):
Changes look good

imabdulbasit (approved):

Cargo.toml
```diff
@@ -85,9 +85,9 @@ prometheus = "0.13"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 snafu = "0.8"
-surf-disco = "0.7"
+surf-disco = "0.6"
```
imabdulbasit, line 88: why do we want to downgrade?

jbearer: We don't really, but we can't merge this into the sequencer with 0.7 until the sequencer is also on 0.7, and *that* can't happen until HotSHot is on 0.7

imabdulbasit: right

imabdulbasit:
ah please bump version

src/data_source/fetching.rs
```diff
@@ -1232,22 +1247,31 @@ where
...
+                        // Wait for the block to be fetched. This slows down the scanner so that we
+                        // don't waste memory generating more active fetch tasks then we can handle
+                        // at a given time. Note that even with this await, all blocks within a
+                        // chunk are fetched in paralle, so this does not block the next block in
+                        // the chunk, only the next chunk until the current chunk completes.
+                        fut.await;
+
+                        // Add a bit of artifical latency to easy pressure on the catchup provider,
+                        // otherwise proactive scans can be highly intensive.
+                        sleep(active_fetch_delay).await;
```
jbearer, line 1272: @imabdulbasit I'm now thinking we should move this sleep (and the one below on line 1288) into the stream itself. E.g. after we [flatten](https://github.com/EspressoSystems/hotshot-query-service/blob/05743e5e5a09d456aced23e8d9a97b970e448a55/src/data_source/fetching.rs#L1007) the stream of chunks into a stream of individual objets, we could then do something like `.then(|obj| async move { sleep(delay).await; obj })` to inject a small delay between each object. This would have the same effect as these sleeps in slowing down proactive fetching, since we are reading from those streams. But it would also eliminate the issue where subscribing to an old stream leads to very high CPU usage because we scan all the old blocks as fast as the server is able. 

In effect, these two things are the same issue: consuming a stream happens too fast.

OTOH, in the more common usage, where your stream has reached the end of the chain and most of your time is blocked waiting for the next block to arrive, you're not going to notice a small delay anyways. The delay just has to be significantly lower than the expected block time so you can easily keep up with the chain. I think 100ms or less is probably good.

Feel free to push to this branch today if you agree.

jbearer:
Offhand I'm guessing the test timeout is from the `NoStorage` tests, which depend heavily on proactive fetching. We already turn the proactive fetching intervals way down for those tests. We probably also have to turn down the new delay to get them to pass.

src/data_source/fetching.rs
jbearer: Or perhaps a delay after each *active fetch* only, as we are doing here, but moved inside the stream itself. Then another delay unconditionally after each chunk. The former limits the requests we make for peers but doesn't slow us down unnecessarily when we already have the data. The latter slows us down a little bit when we are fetching a ton of objects from our own database. Both become negligible when we are passively waiting for new objects to arrive.

imabdulbasit: `But it would also eliminate the issue where subscribing to an old stream leads to very high CPU usage because we scan all the old blocks as fast as the server is able.`

If I am understanding correctly then you mean that we are loading range of blocks from the database constantly and there is no delay between it? If yes then isn't it okay to use max CPU? Maybe we can reduce the chunk size then

imabdulbasit: It can also benefit tho as we don't make concurrent database requests so a delay would provide other requests to be served.

jbearer: Yeah in that case it's technically not a problem to use max CPU...since we're yielding every chunk if we are able to use max CPU it just means everyone else gets to run. It just looks a little scary in Datadog lol

imabdulbasit: 1cdd9b77a203b391871f1923264a1dd285771287
Lmk if these changes look good to you. I want to merge this PR and create a new tag before weekly release merge

imabdulbasit:
> Offhand I'm guessing the test timeout is from the `NoStorage` tests, which depend heavily on proactive fetching. We already turn the proactive fetching intervals way down for those tests. We probably also have to turn down the new delay to get them to pass.

https://github.com/EspressoSystems/hotshot-query-service/pull/603/commits/d9977b6099e6b29b79f44f2011644dd94602b015

src/data_source/fetching.rs
jbearer: The gist of the changes looks good, but I left a couple of suggestions. Somehow the comments went on the commit page and not the PR, you can see them [here](https://github.com/EspressoSystems/hotshot-query-service/commit/1cdd9b77a203b391871f1923264a1dd285771287

PR #618, opened 2024-06-04
Fine tune locking to minimize critical sections

jbearer:
This change should reduce the frequency/severity of slow requests due to long queues forming on the lock protecting the API state.

### This PR:
* Uses `at` instead of `get` to manually control locking in request handlers. The changes are focused on requests which might be slow and which don't actually need a lock on the state for their whole duration
* Upgrades all endpoints in the availability API which have a timeout on a `Fetch` so that they don't have a lock while waiting for the timeout
* Refactors the `sync-status` endpoint to make a clone of the SQL client, and run the actual query without borrowing/locking the API state. This allows the (potentially very slow) query to proceed asynchronously from other requests and state updates

### DIFF
--- a/src/data_source/storage/sql.rs
+++ b/src/data_source/storage/sql.rs
@@ -52,6 +52,7 @@ use bit_vec::BitVec;
 use chrono::Utc;
 use committable::Committable;
 use futures::{
+    future::{BoxFuture, FutureExt},
     stream::{self, BoxStream, StreamExt, TryStreamExt},
     task::{Context, Poll},
     AsyncRead, AsyncWrite,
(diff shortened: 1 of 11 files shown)


src/data_source/storage/sql.rs
```diff
@@ -1059,73 +1060,78 @@ where
...
-        // need to select the total number of VID rows and the number of present VID rows with a
-        // NULL share.
-        let row = self
-            .query_one_static(
-                "SELECT max_height, total_leaves, null_payloads, total_vid, null_vid, pruned_height FROM
-                    (SELECT max(leaf.height) AS max_height, count(*) AS total_leaves FROM leaf),
-                    (SELECT count(*) AS null_payloads FROM payload WHERE data IS NULL),
-                    (SELECT count(*) AS total_vid FROM vid),
-                    (SELECT count(*) AS null_vid FROM vid WHERE share IS NULL),
+    async fn sync_status(&self) -> BoxFuture<'static, QueryResult<SyncStatus>> {
```
imabdulbasit, line 1063: Why are we returning BoxFuture?

src/node.rs
```diff
@@ -135,8 +135,12 @@ where
             }
             .boxed()
         })?
-        .get("sync_status", |_req, state| {
-            async move { Ok(state.sync_status().await?) }.boxed()
+        .at("sync_status", |_req, state| {
+            async move {
+                let fut = state.read(|state| state.sync_status()).await;
```
imabdulbasit, line 140: ok okay. I think its to avoid read lock here

jbearer: Oh yeah, in terms of why are we returning a future at all as opposed to just returning a result, that is why.

imabdulbasit (approved):

PR #622, opened 2024-06-05
Do batch delete and pruned height upsert in one transaction block

imabdulbasit:
- Save pruned height after each batch delete and in same transaction block as batch delete
- adjust default interval to 1.5 hours

### DIFF
--- a/src/data_source/fetching.rs
+++ b/src/data_source/fetching.rs
@@ -370,7 +370,9 @@ where
                         match storage.storage.prune().await {
                             Ok(Some(height)) => {
                                 storage.pruned_height = Some(height);
-                                tracing::warn!("pruner run {i} succeeded. Pruned to height {height}");
+                                tracing::warn!(
+                                    "pruner run {i} succeeded. Pruned to height {height}"
+                                );
                             }
                             Ok(None) => (),
                             Err(e) => {
--- a/src/data_source/fetching.rs
+++ b/src/data_source/fetching.rs
@@ -362,7 +362,7 @@ where
             BackgroundTask::spawn("pruner", async move {
                 for i in 1.. {
                     sleep(cfg.interval()).await;
-                    tracing::warn!("pruner woke up for the {i}th time",);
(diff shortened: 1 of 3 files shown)


src/data_source/fetching.rs
```diff
@@ -362,24 +362,18 @@ where
...
-                                    tracing::error!("failed to save pruned height: {e:?}");
-                                    continue;
-                                }
-                                if let Err(e) = storage.commit().await {
-                                    tracing::error!("failed to commit: {e:?}")
-                                }
                             }
                             Ok(None) => (),
                             Err(e) => {
+                                storage.revert().await;
```
jbearer, line 376: Put this after the `pruning failed` log so if `revert` does anything weird it will be obvious why

src/data_source/fetching.rs
```diff
@@ -362,24 +362,18 @@ where
...
                     sleep(cfg.interval()).await;
-                    tracing::info!("pruner woke up for the {i}th time",);
+                    tracing::warn!("pruner woke up for the {i}th time",);
 
                     {
                         let mut storage = fetcher.storage.write().await;
 
                         match storage.storage.prune().await {
                             Ok(Some(height)) => {
                                 storage.pruned_height = Some(height);
```
jbearer, line 372: Log something here like `pruned to height {height}`

jbearer (commented):
This looks like a good change, and will even allow us in the future to release the lock on storage in between batches if we so choose (although for now I don't think it's a big deal, because we're pruning incrementally each prune should apply to relatively little data and therefore be pretty fast).

Still, I don't directly see how this could cause/fix the issues were seeing in cappuccino. The fact that we aren't doing this in a transaction is definitely concerning, but we do have an exclusive lock on the connection while this is happening, and unless there are any logs indicating that something in pruning actually failed -- idk, can you think of a specific scenario where this could explain the behavior we're seeing?

jbearer (approved):

PR #629, opened 2024-06-06
sleep and drop write lock after each batch delete

imabdulbasit:


src/data_source/fetching.rs
```diff
@@ -358,33 +358,39 @@ where
...
+                        }
+                        Err(e) => {
+                            tracing::error!("pruner run {i} failed: {e:?}");
+                            storage.revert().await;
                         }
                     }
 
-                    sleep(cfg.interval()).await;
+                    drop(storage);
+                    sleep(cfg.batch_delay()).await;
```
jbearer, line 385: tbh I don't think we even need to sleep here. Just releasing the lock is enough, if other tasks want to run then we will be paused, otherwise we might as well move straight to the next batch

src/data_source/fetching.rs
```diff
@@ -358,33 +358,39 @@ where
...
+                            storage.pruned_height = Some(height);
+                            tracing::warn!("Pruned to height {height}");
+                        }
+                        Ok(None) => {
+                            tracing::warn!("pruner run {i} complete.");
+                            break;
+                        }
+                        Err(e) => {
+                            tracing::error!("pruner run {i} failed: {e:?}");
+                            storage.revert().await;
```
jbearer, line 380: Should we break here? It's a tough call because we might succeed on the next batch, but on the other hand we don't want to get into an infinite loop of pruning and failing if it's something more serious. So we could potentially just wait and try again after the next interval

src/data_source/storage/sql.rs
```diff
@@ -361,10 +361,29 @@ impl Config {
...
+    pruned_height: u64,
+    run_in_progress: bool,
+}
+
+impl Pruner {
+    fn new(config: PrunerCfg) -> Self {
+        Pruner {
+            enabled: true,
+            config,
+            pruned_height: 0,
```
jbearer, line 381: Why does `pruned_height` start at 0? Shouldn't it be loaded from the db?

jbearer: Oh, I see, is it because this is only used when `run_in_progress`?

src/data_source/storage/sql.rs
```diff
@@ -600,22 +631,23 @@ impl PruneStorage for SqlStorage {
             .await?;
 
         if let Some(target_height) = target_height {
-            while height < target_height {
+            if height < target_height {
```
jbearer, line 634: I feel like we should save `target_height` in `self.pruner` when a run is in progress, so that all batches within a single "run" are moving towards the same target. If we reload target height each time, there is some risk that we never finish a run, because we return after the first batch, and by the time we are called again, the target height has moved forward just a little. Then we end up pruning really inefficiently one block at a time or so

src/data_source/storage/sql.rs
imabdulbasit: removed run_in_progress field and made pruned_height optional to make it more clear

jbearer (approved):

PR #655, opened 2024-07-01
Improve reliability of TmpDb startup

jbearer:
Because we were only checking for database readiness from inside the container, there was a case where we would consider the TmpDb "ready" and allow connections before the Docker daemon had actually exposed the database port to the host. This change explicitly checks that the port is exposed on the host before returning, with no extra dependencies on the host side.

This fixes an intermittent "Connection Refused" error that @rob-maron and I were seeing while debugging an unrelated sequencer test failure.

### DIFF
--- a/src/data_source/storage/sql.rs
+++ b/src/data_source/storage/sql.rs
@@ -3156,7 +3156,7 @@ pub mod testing {
             // ensure the port is open by just establishing a TCP connection.
             while let Err(err) = TcpStream::connect(format!("{}:{}", self.host, self.port)).await {
                 tracing::warn!("database is ready, but port is not available to host: {err:#}");
-                sleep(Duration::from_secs(1)).await;
+                sleep(Duration::from_millis(100)).await;
             }
         }
     }
--- a/src/data_source/storage/sql.rs
+++ b/src/data_source/storage/sql.rs
@@ -2993,6 +2993,7 @@ fn build_get_path_query(
 // These tests run the `postgres` Docker image, which doesn't work on Windows.
 #[cfg(all(any(test, feature = "testing"), not(target_os = "windows")))]
 pub mod testing {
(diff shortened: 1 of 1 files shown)


src/data_source/storage/sql.rs
```diff
@@ -3147,6 +3148,16 @@ pub mod testing {
...
             }
+
+            // The above command ensures the database is ready inside the Docker container. However,
+            // on some systems, there is a slight delay before the port is exposed via host
+            // networking. We don't need to check again that the database is ready on the host (and
+            // maybe can't, because the host might not have pg_isready installed), but we can
+            // ensure the port is open by just establishing a TCP connection.
+            while let Err(err) = TcpStream::connect(format!("{}:{}", self.host, self.port)).await {
+                tracing::warn!("database is ready, but port is not available to host: {err:#}");
+                sleep(Duration::from_secs(1)).await;
```
rob-maron, line 3159: Should we consider adding a timeout?

rob-maron (approved):

tbro (approved):
My only thought is that 1 second seems like a long time for this sort of thing to occur, so might be worth reducing the wait time.

src/data_source/storage/sql.rs
jbearer: I'm torn. Realistically the only thing we could do after a timeout is panic...so it seems better to keep retrying and let the test runner worry about timeouts. OTOH this could be annoying for local testing. Maybe a timeout that is configurable via an env var so we can easily adjust it if it is just set too short for CI or anything?

jbearer: [be8de83](https://github.com/EspressoSystems/hotshot-query-service/pull/655/commits/be8de83e07327a96d494a708c73336c148445846)

Issue #271, opened 2023-11-27
Consider switching SQL client library to sqlx


jbearer:
https://github.com/EspressoSystems/hotshot-query-service/pull/265#issuecomment-1827454414

Currently, we use a combination of tokio-postgres (for the basic SQL client), refinery (to handle migrations), and some custom glue to get it all to work on the async-std executor. It looks like sqlx does most of this out of the box (natively supports async-std and has its own migrations system), so it might clean things up if we switched to that.

PR #674, opened 2024-07-30
Jb/sqlx

jbearer:
See #271 

### This PR:

Primarily, decouples database transactions from the data source itself, in preparation for switching from tokio-postgres to SQLX.

Other significant changes:
* Data source updates no longer require an exclusive lock, because the updates are made on a transaction, outside of the main data source, and then committed atomically. This has cascading effects particularly in the fetching notifier/notify_storage, and in tests. Should be a huge win for parallelism and reducing worst-case request latency, especially after we switch to connection pools
* Breaks SQL storage into several smaller files/modules. We have wanted to do this for a while, and I found it almost unworkable to deal with such a big factor with so many compile errors in one 5k line file, so I just went ahead and did this. I will comment on the diffs where code has only been moved but not changed.
* Adds read-only transactions, so now all DB accesses (reads and writes) use transactions. This lets us take advantage of postgres's strong transaction isolation features to get strong consistency semantics in the presence of concurrent transactions without using application locks

### This PR does not:

* Switch to sqlx or set up sqlite, but these features should be much more straightforward to do after this is merged.
* Handle concurrent writes in any intelligent way. We will need to address this before merging. **Input wanted.** Basically, we now allow tasks to write to the database (via transactions) while other tasks are reading from it or even writing to it (via concurrent transactions). In general, any combination of a pruner run, async fetching task, decide event handler, merklized state update loop, and read-only query may be running at the same time and competing for access to the same data. Thus, we must ask ourselves what sort of consistency guarantees each database connection has.

  By default, each Postgres transaction or statement (a naked statement is implicitly treated as a transaction as if it were wrapped in `BEGIN ... COMMIT`) has [`READ COMMITTED`](https://www.postgresql.org/docs/current/transaction-iso.html#XACT-READ-COMMITTED) isolation from other transactions/statements. This means that any single statement in this transaction will only see data committed before the start of the statement, or in other words every statement gets a consistent view of the database. However, multiple statements in a transaction may get different views of the database, and thus business logic constraints that we expect to hold for different queries over related data may not hold, if another transaction commits in between two statements. Notably, sub-selects are considered separate statements. Also, if there are any places we make read-only queries in separates statements, these would be treated as separate transactions, and may read data that was modified in between.

  I believe it is very difficult to reason about the correctness of our queries with these semantics, especially because we have a fairly complex schema with many relations between tables, and we routinely make complicated queries with multiple predicates, nested sub-selects, and so on. Thus, my recommendation is to use:
  - The [`SERIALIZABLE`](https://www.postgresql.org/docs/current/transaction-iso.html#XACT-SERIALIZABLE) isolation level for modifying transactions. The main downside of using this strict isolation level is that a `COMMIT` may fail if a competing transaction committed first and there is no way to serialize the two transactions (typically happens when they modify the same data). This requires the transaction to be retried, but we already have retry logic in place everywhere, so this shouldn't be a problem. Moreover, it is actually extremely rare that we do have concurrent writers: the only tasks that frequently write are the event handler task and the merklized state task (which operates on separate data). Other tasks only run rarely (like the pruner and proactive fetcher) or in exceptional circumstances (like active fetch tasks when data is missing). Thus, I expect it will be very rare that a transaction actually reverts due to a serialization error. 
  - The `SERIALIZABLE READ ONLY DEFERRED` isolation level for read-only transactions, explicitly wrapping all read-only queries in `BEGIN`/`END` so we get consistent results even for queries that include multiple statements (like the merklized state `get_path` queries). This has the same strong consistency guarantees as we have for writes, with the added benefit that the transaction will never fail with a serialization error (although it may block initially until a fully consistent snapshot is available to read from). Because writes are relatively infrequent (once every few seconds) I believe it will not be that often that one of these transactions actually has to block.

  These isolation levels give us the strongest possible consistency guarantees for concurrent transactions: the results are always as if the transactions had each executed serially in some arbitrary order. Morever, I believe our application meets all the performance considerations recommended for using this mode (see the last paragraph of [13.2.3](https://www.postgresql.org/docs/current/transaction-iso.html#XACT-SERIALIZABLE)).

  For the sake of completeness, I think there is an argument to be made for using `READ COMMITTED` or `REPEATABLE READ`. Basically, our database modifications typically only add data that wasn't there before (like inserting a new block or state snapshot). We don't modify existing data in place. Thus, it is hard to think of a scenario where consecutive queries reading from two different consistent database snapshots would actually give an invalid result. At the end of the day, though, I just don't think it's worth the stress to try and prove this, when we can just rely on the strong consistency of `SERIALIZABLE` unless/until it presents performance problems for us. Also worth noting that this argument kind of falls apart when you introduce pruning.

  Anyways, I am looking for feedback on wheteher `SERIALIZABLE` does sound like the way to go, and I am also open to alternative approaches.

### Key places to review:

I would look at everything, but I will comment on the diff where you can skip code that only moved locations, but didn't have significant changes. I recommend reviewing in the following "top-down" order:

* Get familiar with the usage of the new [transaction interface](https://github.com/EspressoSystems/hotshot-query-service/compare/main...jb/sqlx#diff-ca2e0f4e560077a5c45cd46210a44e9d0e2ee995e31d99d346d7c883728e94efR163-R192) (and trait bounds), e.g. in [the docs](https://github.com/EspressoSystems/hotshot-query-service/compare/main...jb/sqlx#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R78-R82), [helpers](https://github.com/EspressoSystems/hotshot-query-service/compare/main...jb/sqlx#diff-c585f6c873898e789a45c669d6c216c1703eee0d606a6ac87c2da1b4c61986a5R352-R354), and [tests](https://github.com/EspressoSystems/hotshot-query-service/compare/main...jb/sqlx#diff-7136d6b9f26b655819ccf69416ccecb4d815aa64427a81900de400e450a44f15R462-R513)
* Review the changes to [`NotifyStorage`](https://github.com/EspressoSystems/hotshot-query-service/compare/main...jb/sqlx#diff-93820ec3be7654382851034c97ad872bdde52ffb3dd4369e17983dfa93aa797d) and [`Notifier`](https://github.com/EspressoSystems/hotshot-query-service/compare/main...jb/sqlx#diff-26034378cf1df4ad6c2ca7755a5eea889a958934bcdaf4ae3647ef51b1e699ca), where the synchronization/locking is changed to support concurrent reads/writes and multiple writers as long as the underlying database does
* Look at the rest of the [fetching](https://github.com/EspressoSystems/hotshot-query-service/compare/main...jb/sqlx#diff-c54d622f252b84cb4856f3edab92845462ac1bd981023eda79ff4d1b5a758e8b) module (including sub-modules)
* Finally, review the low-level implementation of the new [transaction](https://github.com/EspressoSystems/hotshot-query-service/compare/main...jb/sqlx#diff-1c79423cfdc4f5820421433e96e5ade7172ec4f88dc7890db6d8ba40558ee70c) logic at the database layer
* You can also review the transaction implementation for the [file system backend](https://github.com/EspressoSystems/hotshot-query-service/compare/main...jb/sqlx#diff-6bbc5f12b9a6ca8e749a42f7e44de731a77d77ab8e9dfbe763eed97a7b866956), although it's less important because we should be getting rid of it very soon
* Everything else should be either cut/pasted code or logical consequences of the above changes


src/data_source/storage/sql.rs
```diff
@@ -215,20 +168,50 @@ fn add_custom_migrations(
 }
 
 /// Postgres client config.
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
 pub struct Config {
-    client_config: client::Config,
```
jbearer, line 220: Removed the resettable client stuff as it was complicated to update with the new design, and will be obsolete anyways once we switch to sqlx, where the transaction pool does its own monitoring and refreshing of connections

src/data_source/storage/sql.rs
```diff
@@ -438,48 +414,39 @@ impl SqlStorage {
...
-            return None;
-        }
-
-        Some(self.pruner.config.clone())
+        self.pruner_cfg.clone()
     }
 }
 
 impl SqlStorage {
-    async fn delete_batch(&mut self, height: u64) -> QueryResult<()> {
```
jbearer, line 472: Mutations are moved to the transaction type in `transaction.rs`

src/data_source/storage/sql.rs
```diff
@@ -511,67 +478,24 @@ impl SqlStorage {
...
-    ///
-    /// This function is similar to `AvailabilityStorage::get_header`, but
-    /// * does not require the `QueryablePayload<Types>` bound that that trait impl does
-    /// * makes it easier to specify types since the type parameter is on the function and not on a
-    ///   trait impl
-    /// * allows type conversions for the `id` parameter
-    ///
-    /// This more ergonomic interface is useful as loading headers is important for many SQL storage
-    /// functions, not just the `AvailabilityStorage` interface.
-    async fn load_header<Types: NodeType>(
```
jbearer, line 525: Moved to `query.rs`

src/data_source/storage/sql.rs
```diff
@@ -583,8 +507,7 @@ impl PruneStorage for SqlStorage {
     /// This is because a vacuum operation may be necessary to reclaim more space.
     /// PostgreSQL already performs auto vacuuming, so we are not including it here
     /// as running a vacuum operation can be resource-intensive.
-
-    async fn prune(&mut self) -> Result<Option<u64>, QueryError> {
+    async fn prune(&self, pruner: &mut Pruner) -> anyhow::Result<Option<u64>> {
```
jbearer, line 510: This is the only mutation still defined on the data source itself, since it opens and commits transactions internally

src/data_source/storage/sql.rs
```diff
@@ -653,2342 +577,135 @@ impl PruneStorage for SqlStorage {
...
-                    sql_param(&common_data),
-                ]],
-            )
-            .await
-        }
-    }
-}
-
-#[async_trait]
-impl<Types> NodeDataSource<Types> for SqlStorage
```
jbearer, line 1036: Moved to `query/node.rs`

imabdulbasit (approved):

src/data_source/fetching/notify_storage.rs
```diff
@@ -0,0 +1,558 @@
...
+where
+    Types: NodeType,
+    Payload<Types>: QueryablePayload<Types>,
+    T: UpdateAvailabilityData<Types> + Send + Sync,
+{
+    async fn insert_leaf(&mut self, leaf: LeafQueryData<Types>) -> anyhow::Result<()> {
+        // Store the new leaf.
+        self.inner.insert_leaf(leaf.clone()).await?;
+        // Queue a notification about the newly received leaf.
+        self.inserted_leaves.push(leaf);
```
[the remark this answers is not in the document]
jbearer: They are dropped along with the transaction whenever it is committed/reverted/dropped. On commit we also iterate through them and send out the notifications [here](https://github.com/EspressoSystems/hotshot-query-service/pull/674/files#diff-93820ec3be7654382851034c97ad872bdde52ffb3dd4369e17983dfa93aa797dR199)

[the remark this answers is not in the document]
imabdulbasit: gotcha

PR #676, opened 2024-08-02
hotshot-rc0.5.70

tbro:
Updates for hotshot-rc-0.5.70

### DIFF
--- a/src/testing/consensus.rs
+++ b/src/testing/consensus.rs
@@ -29,7 +29,7 @@ use futures::{
 use hotshot::{
     traits::implementations::{MasterMap, MemoryNetwork},
     types::{Event, SystemContextHandle},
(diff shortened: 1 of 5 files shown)


src/testing/consensus.rs
```diff
@@ -207,7 +207,12 @@ impl<D: DataSourceLifeCycle + UpdateStatusData> MockNetwork<D> {
                                 .unwrap(),
                             ConsensusMetricsValue::new(&*data_source.populate_metrics()),
                             hs_storage,
-                            TestAuctionResultsProvider::default(),
+                            MarketplaceConfig {
```
imabdulbasit, line 210: this is the only change which we don't have in https://github.com/EspressoSystems/hotshot-query-service/pull/675/files

imabdulbasit (approved):

PR #693, opened 2024-09-11
Ensure dropped transactions are reverted in a timely manner

jbearer:
The old method of queuing dropped transactions for revert and then actually doing the revert the next time the transaction is used in an async context can leave idle transactions unreverted for a long time. This may seem ok if the connection is not otherwise being used. However, this can cause `SERIALIZABLE READ ONLY DEFERRABLE` transactions to block indefinitely waiting to acquire a snapshot, since the database sees that there is a writable transaction in progress indefinitely.

The new method ensures transactions are always reverted soon after they are dropped by running the revert in a background task. If the connection is acquired again before the revert has completed, it will block on the cmpletion of the background task, to ensure the connection is not used while the old transaction is still in progress.

### DIFF
--- a/src/data_source/storage/sql/transaction.rs
+++ b/src/data_source/storage/sql/transaction.rs
@@ -35,6 +35,7 @@ use crate::{
         update::{self, ReadOnly},
     },
     merklized_state::{MerklizedState, UpdateStateData},
+    task::Task,
     types::HeightIndexed,
     Header, Payload,
     QueryError::{self, NotFound},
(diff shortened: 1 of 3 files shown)


Ayiga (approved):
This looks like it should be fine so long as we can guarantee that the `Connection` struct isn't shared, multiple `Transaction`s cannot live for the same `Connection` at once, and that the `Transaction` stays alive at least through the `commit` call.

jbearer:
Yep, the `MutexGuard` ensures that each `Connection` is completely owned by its `Transaction`. And the `Transaction` interface only permits use that commits or reverts the transaction before releasing it

PR #697, opened 2024-09-12
Revert fs transaction on drop and add regression test

jbearer:
The test checks for this bug as well as the earlier fixed bug where dropping a mutable SQL transaction causes the next immutable transaction to block

### DIFF
--- a/src/data_source/storage/fs.rs
+++ b/src/data_source/storage/fs.rs
@@ -270,9 +270,41 @@ where
     }
 }
 
+pub trait Revert {
+    fn revert(&mut self);
+}
+
+impl<'a, Types> Revert for RwLockWriteGuard<'a, FileSystemStorageInner<Types>>
+where
+    Types: NodeType,
+    Payload<Types>: QueryablePayload<Types>,
+{
+    fn revert(&mut self) {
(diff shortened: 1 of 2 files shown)


src/data_source/storage/fs.rs
```diff
@@ -291,12 +323,9 @@ where
...
 
-    fn revert(mut self) -> impl Future + Send {
-        async move {
-            self.inner.leaf_storage.revert_version().unwrap();
-            self.inner.block_storage.revert_version().unwrap();
-            self.inner.vid_storage.revert_version().unwrap();
-        }
+    fn revert(self) -> impl Future + Send {
+        // The revert is handled when `self` is dropped.
+        async move {}
```
imabdulbasit, line 328: I think we should include the revert here as well. If this function is called anywhere, it's expected to trigger a revert

imabdulbasit (approved):

src/data_source/storage/fs.rs
jbearer: Sorry, comment is not clear. `self` is dropped in this function, so without doing anything here, it does trigger a revert (if we reverted explicitly it would revert twice)

PR #700, opened 2024-09-21
[WEEKLY RELEASE] HotShot - rc-0.5.75

shenkeyao:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/src/availability/query_data.rs
+++ b/src/availability/query_data.rs
@@ -18,7 +18,7 @@ use hotshot_types::{
     traits::{
         self,
         block_contents::{BlockHeader, GENESIS_VID_NUM_STORAGE_NODES},
-        node_implementation::NodeType,
+        node_implementation::{NodeType, Versions},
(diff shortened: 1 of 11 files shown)


src/availability/query_data.rs
```diff
@@ -213,23 +213,25 @@ impl<Types: NodeType> LeafQueryData<Types> {
         leaf: Leaf<Types>,
         qc: QuorumCertificate<Types>,
     ) -> Result<Self, InconsistentLeafError<Types>> {
+        // TODO: Replace with the new `commit`.
+        let leaf_commit = <Leaf<Types> as Committable>::commit(&leaf);
```
imabdulbasit, line 217: What is the new function that we need to call for the new commit? I think some more comments for that would be great

shenkeyao: It's just `leaf.commit`. I can add a more detailed comment.

PR #705, opened 2024-09-25
Add Option values to Explorer histogram data

### LINKED ISSUE #703: Allow for nullability in retrieved histogram data
There are points when loading the Block Explorer's Explorer Page where the endpoint results in an error.  This error is caused by the information in a block's header being stored before the block's payload.  A query for this information can occur between these two writes, resulting in the inability to find a payload for a given block.  When this occurs, the ExplorerHistogram population process via SQL fails as we aren't expecting a `null` value when retrieving this data.

The proposed solution is to allow for some `null` values to present in the `ExplorerHistograms` members:

https://github.com/EspressoSystems/hotshot-query-service/blob/426217cb8d0efb28b914f563dfd0e5f7df0812a6/src/data_source/storage/sql/query/explorer.rs#L407
https://github.com/EspressoSystems/hotshot-query-service/blob/426217cb8d0efb28b914f563dfd0e5f7df0812a6/src/explorer/query_data.rs#L472

Ayiga:
Closes #703

### This PR:

Updates the `ExplorerHistogram` fields for both `time` and `size` to have `Option` values, which can result in a `null` value coming down to the client.

### DIFF
--- a/src/data_source/storage/sql/query/explorer.rs
+++ b/src/data_source/storage/sql/query/explorer.rs
@@ -404,30 +404,12 @@ where
                         let time: Option<i64> = row.try_get("time").map_err(|e| QueryError::Error {
                             message: format!("failed to get column time {e}"),
                         })?;
-                        let size: i32 = row.try_get("size").map_err(|e| QueryError::Error {
+                        let size: Option<i32> = row.try_get("size").map_err(|e| QueryError::Error {
                             message: format!("failed to get column size {e}"),
                         })?;
                         let num_transactions: i64 =
                             row.try_get("transactions").map_err(|e| QueryError::Error {
                                 message: format!("failed to get column transactions {e}"),
-                            })?;
-
-                        let height: u64 = height.try_into().map_err(|e| QueryError::Error {
-                            message: format!("failed to convert height to u64 {e}"),
-                        })?;
-                        let timestamp: u64 =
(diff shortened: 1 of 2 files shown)


alysiahuggins (approved):

src/data_source/storage/sql/query/explorer.rs
```diff
@@ -404,32 +404,14 @@ where
                         let time: Option<i64> = row.try_get("time").map_err(|e| QueryError::Error {
                             message: format!("failed to get column time {e}"),
                         })?;
-                        let size: i32 = row.try_get("size").map_err(|e| QueryError::Error {
+                        let size: Option<i32> = row.try_get("size").map_err(|e| QueryError::Error {
```
imabdulbasit, line 407: payload.size is a required field in schema so is it ever going to return None? I think we have to do left join above?

imabdulbasit: nvm its nullable

imabdulbasit (approved):

src/data_source/storage/sql/query/explorer.rs
Ayiga: I actually this is this probably still a valid concern.

imabdulbasit: we insert header and null payload in insert_leaf() function so no need for left join here

PR #711, opened 2024-10-22
Add metrics and tracing logging around SQL transactions

jbearer:
If the sqlx changes don't fix the issues we're seeing in devnet, this should help with debugging.

### This PR:
Adds the following metrics:
* `sql_open_transactions`: number of transactions currently open
* `transaction_duration`: histogram of transaction durations
* `committed_transactions`: count of transactions closed by committing
* `reverted_transactions`: count of transactions closed by reverting
* `dropped_transactions`: count of transactions closed by dropping (implicit revert)

Adds trace-level logging for transaction lifecycle

### DIFF
--- a/src/data_source/storage/sql/transaction.rs
+++ b/src/data_source/storage/sql/transaction.rs
@@ -43,7 +43,10 @@ use committable::Committable;
 use derive_more::{Deref, DerefMut};
 use futures::{future::Future, stream::TryStreamExt};
 use hotshot_types::traits::{
-    block_contents::BlockHeader, node_implementation::NodeType, EncodeBytes,
+    block_contents::BlockHeader,
+    metrics::{Counter, Gauge, Histogram, Metrics},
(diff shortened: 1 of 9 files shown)


src/data_source/storage/sql/transaction.rs
```diff
@@ -99,6 +108,59 @@ impl TransactionMode for Read {
...
+
+        Self {
+            started_at,
+            metrics,
+            close_type: CloseType::Drop,
+            _mode: Default::default(),
+        }
+    }
+
+    fn close(&mut self, t: CloseType) {
```
shenkeyao, line 146: Nit: Am I understanding correctly that this function sets the close type but it's the `drop` function that does the real closing operation? Can we rename this to `close_type` or `set_close`?

shenkeyao (approved):
Only a nit comment.

shenkeyao (approved):

PR #712, opened 2024-10-23
Add an option for setting the threshold at which statements are considered slow

jbearer:


src/data_source/storage/sql.rs
```diff
@@ -174,11 +175,8 @@ fn add_custom_migrations(
 
 /// Postgres client config.
 #[derive(Clone, Debug)]
-pub struct Config<DB>
-where
-    DB: Database,
-{
-    db_opt: <DB::Connection as Connection>::Options,
+pub struct Config {
+    db_opt: PgConnectOptions,
```
imabdulbasit, line 179: So we would need two Configs when we make `Sqlite` change with that feature thing?

imabdulbasit: Okay, I will revert this change then in Sqlite PR

src/data_source/storage/sql.rs
```diff
@@ -364,6 +360,16 @@ impl<DB: Database> Config<DB> {
         self.pool_opt = self.pool_opt.max_connections(max);
         self
     }
+
+    /// Log at WARN level any time a SQL statement takes longer than `threshold`.
+    ///
+    /// The default threshold is 1s.
+    pub fn slow_statement_threshold(mut self, threshold: Duration) -> Self {
```
imabdulbasit, line 367: do we want to set this at sequencer side through an env variable?

imabdulbasit (approved):

src/data_source/storage/sql.rs
jbearer: probably yeah

PR #716, opened 2024-10-25
Slow down active fetching task

jbearer:
By awaiting the objects we are fetching, we ensure that we do not spawn active fetching tasks at a faster rate than they complete, thus preventing a resource leak.

### DIFF
--- a/src/data_source/fetching.rs
+++ b/src/data_source/fetching.rs
@@ -995,7 +995,17 @@ where
                         chunk_size,
                         start..block_height,
                     );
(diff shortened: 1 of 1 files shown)


jbearer:
At one point in the past we were doing this. Somehow it got changed, so that we were just dropping the inner future. The fetch would still run in a spawned task, but this allowed the scanner task to get way ahead and accumulate many spawned tasks

rob-maron:
Yeah I remember this in the past

PR #722, opened 2024-10-29
Replace explorer prefix search with exact match for block and transaction hash

imabdulbasit:
addresses https://github.com/EspressoSystems/espresso-block-explorer/issues/148

This PR:

- replaces the prefix search with equality for block and transaction searches
- replace search query parameter type from `String` to `TaggedBase64`
- queries database for blocks or transactions based on the tag

### DIFF
--- a/src/data_source/storage/sql/queries/explorer.rs
+++ b/src/data_source/storage/sql/queries/explorer.rs
@@ -519,12 +519,11 @@ where
         let header_tag = Commitment::<Header<Types>>::tag();
         let tx_tag = Commitment::<HotshotTransaction<Types>>::tag();
 
-        let search_query_string = search_query.to_string();
-
         if search_tag != header_tag && search_tag != tx_tag {
             return Err(GetSearchResultsError::InvalidQuery(errors::BadQuery {}));
         }
 
+        let search_query_string = search_query.to_string();
         if search_tag == header_tag {
             let block_query = format!(
                 "SELECT {BLOCK_COLUMNS}
--- a/src/data_source/storage/sql/queries/explorer.rs
+++ b/src/data_source/storage/sql/queries/explorer.rs
@@ -531,18 +531,17 @@ where
                     JOIN payload AS p ON h.height = p.height
                     WHERE h.hash = $1
                     ORDER BY h.height DESC
-                    LIMIT 5"
+                    LIMIT 1"
             );
-            let block_query_rows = query(block_query.as_str())
+            let row = query(block_query.as_str())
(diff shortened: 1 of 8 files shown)


src/data_source/storage/sql/queries/explorer.rs
```diff
@@ -509,61 +513,76 @@ where
...
+                "SELECT {BLOCK_COLUMNS}
+                    FROM header AS h
+                    JOIN payload AS p ON h.height = p.height
+                    WHERE h.hash = $1
+                    ORDER BY h.height DESC
+                    LIMIT 5"
+            );
+            let block_query_rows = query(block_query.as_str())
+                .bind(&search_query_string)
+                .fetch(self.as_mut());
```
jbearer, line 538: We can do `LIMIT 1` and `fetch_one` here, querying for blocks by equality on hash is only ever going to return one block

jbearer (approved):
@Ayiga can you confirm this will work with the block explorer?

src/data_source/storage/sql/queries/explorer.rs
imabdulbasit: I am considering making the blocks field an Option<BlockSummary> since there can only be one block now. However, I don't want to break the frontend, and I also think that SearchResult should be an enum

Ayiga:
> @Ayiga can you confirm this will work with the block explorer?

I think this should be fine in terms of functionality at the moment. There is a slight change in the input requirements that may cause problems with the current version of the Block Explorer.  That being the requirement of the input to be a `TaggedBase64`.  Though I do think that this is a good change, it does mean that the unchanged search request will start returning `4xx` errors where it before didn't.

But this is likely alright, as we'll want to change the Block Explorer's search behavior regardless so that it isn't submitting these requests when they don't conform to the `TaggedBase64` requirements.

src/data_source/storage/sql/queries/explorer.rs
Ayiga: Making this return an `enum` would indeed be a backwards incompatible change.  I can possibly support it with a backwards compatible decoder that can support either representation, but I think that would need to be in place in the Block Explorer client-side code before that sort of change is introduced on the server side.

Ayiga (approved):

Ayiga:
Just tested this.  The `4xx` errors aren't affecting the current block explorer in any meaningful way.  So this should be relatively safe.  Though, gain we'd want to update the logic in the block explorer to not send partial `TaggedBase64` search terms.

Issue #721, opened 2024-10-29
Explorer API Transaction indexing is reversed


Ayiga:
When navigating transactions using the block explorer, the order in which we visit / link transactions is reversed.

We identify an individual transaction via its `block height` and its `transaction offset`.  This transaction offset is always meant to be in ascending order, such that the `0` offset is the first transaction within the block.

This gets a little weird when it comes to retrieving the list of "Latest Transactions" as we want that list to be descending overall, but still return the appropriate offset number correctly.

Where this gets odd is in retrieving the `Transaction Detail` for a given `(height, offset)` pair.  Instead of returning the transaction that corresponds to the given `(height, offset)` we instead end up returning the transaction that corresponds to `(height, N - offset)` where `N` is the total number of transactions within the block at `height`.

Example:
https://explorer.decaf.testnet.espresso.network/block/315286

This block has two transactions within it.  If we look at its transaction page:
https://explorer.decaf.testnet.espresso.network/transactions?block=315286

We can see both of these transactions.  We can even see that they list links to these pages:
- https://explorer.decaf.testnet.espresso.network/transaction/315286-1
- https://explorer.decaf.testnet.espresso.network/transaction/315286-0

But notice that the hashes of each of the clicked links doesn't match the initially listed hash.

`TX~v_HZvxLYwnnqkSqr56edsFVuXdnXs5nJM3JkS1TAwg1N` indicates that it links to https://explorer.decaf.testnet.espresso.network/transaction/315286-1, but that page shows `TX~Nev36VRr3zOQ-BFRZICxG-aLzDQoe1nlS24G4PxWBA8R` instead.

This is also true for the other.

Ayiga:
This has been resolved with #723.

PR #723, opened 2024-10-29
fix order by clause in getTransactionDetail query

imabdulbasit:
addresses https://github.com/EspressoSystems/hotshot-query-service/issues/721

The PR fixes the order by clause for transaction detail query. 

using ORDER BY DESC with an offset caused skipping of recent transactions, resulting in (N - offset)th transaction. This PR changes DESC to ASC  so that older transactions are correctly skipped


Ayiga (approved):
The `rev` did the trick!

PR #725, opened 2024-10-31
Make `update` fallible again

jbearer:
A recent commit made `update` infallible, reasoning that if we fail to store a new leaf, it is better to at least try storing the next one, rather than failing completely. However, this is not accurate: the client uses the success/failure of `udpate` to determine when it is safe to garbage collect data from temporary storage. Thus, returning success when we didn't actually store all the leaves can lead to data loss.

This change fixes the situation by again returning an error whenever we fail to insert a leaf. It also tries to make the return value more useful, by returning on error the height of the first leaf we failed to insert (which is useful for garbage collection) instead of just an opaque error message which the caller can't do much with.

### DIFF
--- a/src/data_source/sql.rs
+++ b/src/data_source/sql.rs
@@ -267,7 +267,9 @@ impl Config {
 ///     spawn(async move {
 ///         let mut events = hotshot.event_stream();
 ///         while let Some(event) = events.next().await {
-///             state.hotshot_qs.update(&event).await;
+///             if state.hotshot_qs.update(&event).await.is_err() {
+///                 continue;
+///             }
 ///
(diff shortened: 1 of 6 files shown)


src/data_source/sql.rs
```diff
@@ -334,7 +336,7 @@ pub mod testing {
         }
 
         async fn handle_event(&self, event: &Event<MockTypes>) {
-            self.update(event).await;
+            self.update(event).await.unwrap();
```
imabdulbasit, line 339: why is unwrapping here okay?

imabdulbasit (approved):
LGTM!

PR #726, opened 2024-10-31
Fetch leaves recursively by hash, verifying chaining

### LINKED ISSUE #620: Sync from newest to oldest
It would be better to run proactive scans, especially major scans, backwards, because the newest data is both the most likely to be missing (in the case where a node has just been offline for a short time, like an update) and the most likely to be queried.

Adding a reverse block stream might help simplify some of the transaction streaming stuff for the explorer as well.

jbearer:
This changes leaf fetching to require already having the _next_ leaf. This tells us what the hash should be for the leaf being fetched.

This addresses two unrelated issues:
1. Leaf fetching is trusted: we were not previously verifying that the leaf returned by a peer is valid in any way. Now, we can verify the fetched leaf against the expected hash (and similarly for the fetched QC). We exploit the chaining property of HotShot leaves to avoid having to run any kind of consensus light client to verify fetched leaves.
2. Fetching leaves by hash instead of (or in addition to) by height allows us to implement more providers. For example, we can now implement a provider that pulls leaves from undecided consensus storage, by hash, which allows us to fetch a leaf from our own storage even if we missed the corresponding decide event.

### This PR:
* Changes leaf fetching to require the next leaf, and fetch/authenticate by hash
* Proactive scanning now runs backwards, since leaf fetching becomes kind of inherently backwards. This was a change we wanted anyways (closes #620)
* To facilitate that, we have added reverse streams for all the resource types, which may come in handy for the explorer
* Receiving a leaf (either by fetching or decide event) now triggers us to fetch the parent if it is missing. This is supplements the proactive fetcher and can often help us obtain missing data really quickly (e.g. if we just missed one decide)
* `NoStorage` is gone. The purpose of `NoStorage` was to stress test fetching, which it did in the beginning, but it has been ages since we found a real bug this way; we now have plenty of other adversarial fetching tests, and `NoStorage` is becoming more trouble than it's worth to maintain. In particular, effective fetching now depends on having somewhat reliable storage (a reasonable assumption!) because we assume if you fetch a leaf, you can look it up shortly thereafter and thus fetch its parent. Thus, the `NoStorage` tests were failing or very slow because of many failed requests, with this change.

### Key places to review:
* New leaf fetching logic: `active_fetch` in `src/data_source/fetching/leaf.rs`
* Triggering fetch of the parent leaf: `trigger_fetch_for_parent` in `src/data_source/fetching/leaf.rs`
* Reverse streams: `src/data_source/fetching.rs`


imabdulbasit (approved):
Looks very good! I like how we are using `parent_commitment()` to verify the fetched leaf.

PR #728, opened 2024-11-06
ab/sqlite-support

imabdulbasit:
This PR enables SQLite support for the query service. Since SQLx supports both PostgreSQL and SQLite, most of the code has been reused. An additional feature, `embedded-db`, has been introduced. When enabled, it activates the SQLite code; if not enabled, the PostgreSQL code is used.

The migrations for SQLite and PostgreSQL are separate, as some data types differ between the two. For example, SQLite does not support the `bytea` data type, so we use the `Blob` data type instead, whereas PostgreSQL uses `BYTEA`. Another example is of `SERIAL` data type which SQlite does not support and has `INTEGER PRIMARY KEY AUTOINCREMENT` in replacement

Migrations for SQLite are in` ./migrations/sqlite`, whereas migrations for PostgreSQL have been moved to` ./migrations/postgres.`

This PR also updates the CI build workflow to add another job with only the `no-storage` and `testing` features, which enables PostgreSQL and consequently disables the SQlite 

The follow-up work on the sequencer side will also include a migration for the merklized state tables, as the migrations are in the sequencer repo. This will involve migrating the children_bitvec column from the` Bit varying (BIT(n)) `type to `BYTEA`/`JSONB`, since SQLite does not support bit strings. However, we need to convert the Bit(n) to bytes before storing it, rather than directly converting it to JSON, as that would take 1 byte for each bit in the bit string.

This change also requires the sequencer binary to be compiled separately for SQLite with the embedded feature for the query service.


.github/workflows/build.yml
```diff
@@ -42,13 +42,19 @@ jobs:
       # don't compile with all combinations of features.
       - name: Clippy
         run: cargo clippy --workspace --all-features --all-targets -- -D warnings
-      
-      - name: Test
+
+      - name: Test with embedded-db
```
jbearer, line 46: Would be nice to make these separate jobs rather than separate steps so they could run in parallel

examples/simple-server.rs
```diff
@@ -52,7 +52,7 @@ use std::{num::NonZeroUsize, str::FromStr, time::Duration};
 use url::Url;
 use vbs::version::StaticVersionType;
 
-const NUM_NODES: usize = 2;
+const NUM_NODES: usize = 5;
```
jbearer, line 55: Why?

src/data_source/storage/sql.rs
```diff
@@ -71,15 +78,30 @@ use self::{migrate::Migrator, transaction::PoolMetrics};
...
+///
+/// // For SQLite
+/// #[cfg(feature = "embedded-db")]
+/// {
+///     let mut sqlite_migrations: Vec<Migration> =
+///     include_migrations!("$CARGO_MANIFEST_DIR/migrations/sqlite").collect();
+///     sqlite_migrations.sort();
+///     assert_eq!(sqlite_migrations[0].version(), 10);
+///     assert_eq!(sqlite_migrations[0].name(), "init_schema");
+/// }
```
jbearer, line 104: Nit: this could be condensed
```rust
// For PostgreSQL
#[cfg(not(feature = "embedded-db"))]
let mut migrations: Vec<Migration> = include_migrations!("$CARGO_MANIFEST_DIR/migrations/postgres").collect();

// For SQLite
#[cfg(feature = "embedded-db")]
let mut migrations: Vec<Migration> = include_migrations!("$CARGO_MANIFEST_DIR/migrations/sqlite").collect();

migrations.sort();
assert_eq!(migrations[0].version(), 10);
assert_eq!(migrations[0].name(), "init_schema");
```

src/data_source/storage/sql.rs
```diff
@@ -184,29 +212,64 @@ fn add_custom_migrations(
...
 
-/// Postgres client config.
-#[derive(Clone, Debug)]
-pub struct Config {
-    db_opt: PgConnectOptions,
+pub struct Config<DB>
+where
+    DB: Database,
+{
+    db_opt: <DB::Connection as Connection>::Options,
```
jbearer, line 219: Is there still an advantage to making this generic over DB? Vs like

```rust
#[cfg(feature = "embedded-db")]
db_opt: SqliteConnectOptions,
#[cfg(not(feature = "embedded-db"))]
db_opt: PostgresConnectOptions
```

src/data_source/storage/sql.rs
```diff
@@ -184,29 +212,64 @@ fn add_custom_migrations(
...
     pool_opt: PoolOptions<Db>,
+    #[cfg(not(feature = "embedded-db"))]
     schema: String,
+    #[cfg(feature = "embedded-db")]
+    db_path: std::path::PathBuf,
```
jbearer, line 224: This is part of `db_opt` already, `SqliteConnectOptions` has `filename`

src/data_source/storage/sql.rs
```diff
@@ -184,29 +212,64 @@ fn add_custom_migrations(
...
 }
 
-impl From<PgConnectOptions> for Config {
+#[cfg(feature = "embedded-db")]
+impl Default for Config<sqlx::Sqlite> {
+    fn default() -> Self {
+        SqliteConnectOptions::default()
+            .busy_timeout(Duration::from_secs(30))
+            .auto_vacuum(sqlx::sqlite::SqliteAutoVacuum::Incremental)
+            .into()
```
jbearer, line 250: Should we add `create_if_missing` to the default? Seems like a good default behavior to have

src/data_source/storage/sql.rs
```diff
@@ -401,31 +490,42 @@ pub struct Pruner {
...
+            }
+            .boxed()
+        });
+
+        #[cfg(feature = "embedded-db")]
+        if config.reset {
+            std::fs::remove_file(config.db_path)?;
+        }
+
+        let pool = pool.connect(config.db_opt.to_url_lossy().as_ref()).await?;
```
jbearer, line 515: Why did we change this back to `to_url_lossy`? Some of the options we care about, such as log levels, don't get encoded in the URL, but do get passed through if we use `connect_with(config.db_opt)`

src/data_source/storage/sql.rs
```diff
@@ -672,38 +793,71 @@ impl VersionedDataSource for SqlStorage {
...
     #[derive(Debug)]
     pub struct TmpDb {
+        #[cfg(not(feature = "embedded-db"))]
         host: String,
+        #[cfg(not(feature = "embedded-db"))]
         port: u16,
+        #[cfg(not(feature = "embedded-db"))]
         container_id: String,
+        #[cfg(feature = "embedded-db")]
+        db_path: std::path::PathBuf,
```
jbearer, line 825: Why not use `Tempfile` here? It automatically handles the random name generation, automatically deletes the file on `Drop`, and it creates the files in the user's `/tmp` dir which is nicer than under the working directory

src/data_source/storage/sql.rs
```diff
@@ -672,38 +793,71 @@ impl VersionedDataSource for SqlStorage {
...
+        #[cfg(not(feature = "embedded-db"))]
         persistent: bool,
     }
     impl TmpDb {
+        #[cfg(feature = "embedded-db")]
+        pub fn init_sqlite_db() -> Self {
```
jbearer, line 831: Does this need to be `pub`? Shouldn't users just call `init`

src/data_source/storage/sql.rs
```diff
@@ -672,38 +793,71 @@ impl VersionedDataSource for SqlStorage {
...
+        #[cfg(not(feature = "embedded-db"))]
         host: String,
+        #[cfg(not(feature = "embedded-db"))]
         port: u16,
+        #[cfg(not(feature = "embedded-db"))]
         container_id: String,
+        #[cfg(feature = "embedded-db")]
+        db_path: std::path::PathBuf,
+        #[cfg(not(feature = "embedded-db"))]
         persistent: bool,
```
jbearer, line 827: We could implement this behavior for sqlite too. It would just mean not deleting the file on drop. This is nice sometimes for debugging because the DB is still there after the test fails, so you can look into it

src/data_source/storage/sql.rs
```diff
@@ -931,21 +1135,21 @@ mod test {
...
-                .host(host)
-                .port(port)
-                .migrations(custom_migrations);
-            if !migrations {
-                cfg = cfg.no_migrations();
+        let connect = |migrations: bool, custom_migrations| {
+            let db_opt = db_opt.clone();
+            async move {
+                {
+                    let cfg: Config<Db> = db_opt.into();
```
jbearer, line 1144: Previously we pulled `db_opt` out of `Config` and now we are converting back into `Config`. We could just use `db.config()` directly right?

src/data_source/storage/sql/db.rs
```diff
@@ -37,4 +37,8 @@
 /// {}
```
jbearer, line 37: I think this big long comment is a bit out of date now that we found a better way of doing it

src/data_source/storage/sql/queries.rs
```diff
@@ -94,7 +94,12 @@ impl<'q> QueryBuilder<'q> {
         self.arguments.add(arg).map_err(|err| QueryError::Error {
             message: format!("{err:#}"),
         })?;
-        Ok(format!("${}", self.arguments.len()))
+
+        if cfg!(feature = "embedded-db") {
+            Ok("?".to_string())
```
jbearer, line 99: I believe sqlite does allow the `$` syntax in query params. I don't know if it respects the indexing or just converts it to `?` internally, but I think we might not need this `if`

src/data_source/storage/sql/queries/availability.rs
```diff
@@ -221,12 +221,12 @@ where
...
         let sql = format!(
-            "SELECT {BLOCK_COLUMNS}, t.index AS tx_index
+            "SELECT {BLOCK_COLUMNS}, t.idx AS tx_index
                 FROM header AS h
                 JOIN payload AS p ON h.height = p.height
-                JOIN transaction AS t ON t.block_height = h.height
+                JOIN transactions AS t ON t.block_height = h.height
                 WHERE t.hash = {hash_param}
-                ORDER BY (t.block_height, t.index) ASC
+                ORDER BY t.block_height ASC, t.idx ASC
```
jbearer, line 229: Nit: I think `ASC` is the default anyways so maybe we can just say `t.block_height, t.idx`?

src/data_source/storage/sql/queries/explorer.rs
```diff
@@ -238,7 +238,8 @@ where
         };
 
         let block_height = transaction_target.get::<i64, _>("height") as usize;
-        let transaction_index = transaction_target.get::<Json<TransactionIndex<Types>>, _>("index");
+        let transaction_index =
+            transaction_target.get_unchecked::<Json<TransactionIndex<Types>>, _>("index");
```
jbearer, line 242: Why did this change to unchecked?

src/data_source/storage/sql/queries/node.rs
```diff
@@ -117,7 +117,7 @@ where
                 (SELECT count(*) AS null_payloads FROM payload WHERE data IS NULL) AS p,
                 (SELECT count(*) AS total_vid FROM vid) AS v,
                 (SELECT count(*) AS null_vid FROM vid WHERE share IS NULL) AS vn,
-                coalesce((SELECT last_height FROM pruned_height ORDER BY id DESC LIMIT 1)) as pruned_height
+                (SELECT(SELECT last_height FROM pruned_height ORDER BY id DESC LIMIT 1) as pruned_height)
```
jbearer, line 120: What is happening here?

src/data_source/storage/sql/queries/state.rs
```diff
@@ -321,13 +342,37 @@ pub(crate) fn build_hash_batch_insert(
...
+    R: Row,
+    JsonValue: 'a + Decode<'a, <R as Row>::Database>,
+    Option<JsonValue>: Decode<'a, <R as Row>::Database>,
+    Vec<u8>: Decode<'a, <R as Row>::Database>,
+    i64: Decode<'a, <R as Row>::Database>,
+    i32: Decode<'a, <R as Row>::Database>,
+
+    &'a str: ColumnIndex<R>,
+{
+    fn from(row: &'a R) -> Self {
```
jbearer, line 365: Why did we add this? Doesn't `FromRow` still work?

src/data_source/storage/sql/queries/state.rs
```diff
@@ -58,36 +61,43 @@ where
         // Order by pos DESC is to return nodes from the leaf to the root
 
         let (query, sql) = build_get_path_query(state_type, traversal_path.clone(), created)?;
-        let nodes = query
-            .query_as::<Node>(&sql)
-            .fetch_all(self.as_mut())
-            .await?;
+        let rows = query.query(&sql).fetch_all(self.as_mut()).await?;
+
+        let nodes: Vec<Node> = rows.into_iter().map(|r| (&r).into()).collect();
```
jbearer, line 66: Why not `query_as` anymore?

src/data_source/storage/sql/queries/state.rs
```diff
@@ -371,23 +416,34 @@ fn build_get_path_query<'q>(
...
     let mut sql: String = sub_queries.join(" UNION ");
-    sql.push_str("ORDER BY path DESC");
+
+    sql = format!("SELECT * FROM ({sql}) as t ");
+
+    if cfg!(feature = "embedded-db") {
+        sql.push_str("ORDER BY length(t.path) DESC");
+    } else {
+        sql.push_str("ORDER BY t.path DESC");
+    }
```
jbearer, line 445: I gather this is because sqlite doesn't support ordering on JSONB? However, logically, doesn't ordering by the length do what we want anyways? So maybe it's simpler just to go with that in both backends

src/data_source/storage/sql/transaction.rs
```diff
@@ -274,117 +283,37 @@ impl_tuple_params!(6, (T1, T2, T3, T4, T5, T6,));
 impl_tuple_params!(7, (T1, T2, T3, T4, T5, T6, T7,));
 impl_tuple_params!(8, (T1, T2, T3, T4, T5, T6, T7, T8,));
 
-impl<'p, T> Params<'p> for Vec<T>
+pub fn build_where_in<'a, I>(
+    column: &'a str,
+    query: &'a str,
+    values: I,
```
jbearer, line 289: Nit: it makes more sense to me if we shuffle these arguments so it's `query`, `column`, `values`, because that's the order you would write it in if you were writing out the SQL: `SELECT stuff WHERE {column} IN {values}`

src/testing/consensus.rs
```diff
@@ -66,7 +66,7 @@ pub struct MockNetwork<D: DataSourceLifeCycle> {
 pub type MockDataSource = FileSystemDataSource<MockTypes, NoFetching>;
 pub type MockSqlDataSource = SqlDataSource<MockTypes, NoFetching>;
 
-pub const NUM_NODES: usize = 2;
+pub const NUM_NODES: usize = 5;
```
jbearer, line 69: Why this change?

jbearer (commented):
Lots of minor comments but overall this looks really good! Glad it worked out not *too* painfully haha

examples/simple-server.rs
imabdulbasit: some tests were stuck in View sync so this was an attempt to fix those. I think it did fix it or maybe they are flaky

src/data_source/storage/sql.rs
imabdulbasit: yeah, I think there is also a bug with Drop implementation for posgres.

src/data_source/storage/sql.rs
imabdulbasit: there was an ownership error w this so I did this

src/data_source/storage/sql/queries/node.rs
imabdulbasit: syntax error with sqlite here so had to replace the coalesce here.

src/data_source/storage/sql/queries/explorer.rs
imabdulbasit: so the type checking thing is broken for sqlite. I forgot the exact error but it was something like expected sql type `TEXT` found `INT` or viceversa

src/data_source/storage/sql/queries/state.rs
imabdulbasit: yeah the FromRow does type checking which is broken for sqlite

src/data_source/storage/sql/queries/state.rs
imabdulbasit: same here

src/data_source/storage/sql/queries/state.rs
imabdulbasit: Yeah, postgres has array_length() and sqlite has length() so still need to match the feature

src/data_source/storage/sql/queries/availability.rs
imabdulbasit: yeah I thought it is better for readability but I'll revert

src/data_source/storage/sql.rs
imabdulbasit: didn't click in my mind but don't we want to keep the file in case `persistent` is true

src/data_source/storage/sql/queries.rs
imabdulbasit: you are right!

src/data_source/storage/sql.rs
jbearer: I feel like you could just clone the `Config` and move it into this future but idk, this is fine I guess

src/data_source/storage/sql/queries/explorer.rs
jbearer: Welp

src/data_source/storage/sql/queries/node.rs
jbearer: What is the double select doing?

src/data_source/storage/sql/queries/state.rs
jbearer: Ah right. We should not derive `FromRow` then if it doesn't work

src/data_source/storage/sql/queries/state.rs
jbearer: damn

src/data_source/storage/sql.rs
imabdulbasit: I think I couldn't derive Clone because of generic parameter

src/data_source/storage/sql/queries/node.rs
imabdulbasit: the outer select will return NULL if there are no rows returned by the inner query

jbearer: Ah ok

.github/workflows/build.yml
```diff
@@ -61,3 +56,40 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: ./target/doc
           cname: tide-disco.docs.espressosys.com
+  test-sqlite:
+    runs-on: ubuntu-latest
+    env:
+      RUSTFLAGS: '--cfg async_executor_impl="async-std" --cfg async_channel_impl="async-std"'
+      RUST_LOG: info
+    needs: [build]
```
jbearer, line 64: Does this actually depend on `build`? It checks out and builds things itself so could run in parallel no?

src/data_source/storage/sql.rs
```diff
@@ -672,38 +782,80 @@ impl VersionedDataSource for SqlStorage {
...
+                let (_, db_path) = file.keep().unwrap();
+
+                return Self {
+                    db_path,
+                    persistent,
+                };
+            }
+
+            Self {
+                db_path: file.into_temp_path().to_path_buf(),
```
jbearer, line 837: I think the `TempPath` ends up getting dropped here, which should delete the file. I thought we would need to store the `TempPath` object in `Self` so that the file doesn't get deleted until `Self` is dropped. 

According to [the docs](https://docs.rs/tempfile/latest/tempfile/struct.TempPath.html) there is a way (`keep(bool)`) to construct a `TempPath` such that it does or doesn't get deleted on drop based on a runtime parameter, maybe we can make use of that somehow

jbearer (approved):

jbearer (approved):

src/data_source/storage/sql.rs
```diff
@@ -235,6 +236,8 @@ impl Default for Config {
 impl Default for Config {
     fn default() -> Self {
         SqliteConnectOptions::default()
+            .journal_mode(sqlx::sqlite::SqliteJournalMode::Wal)
+            .synchronous(sqlx::sqlite::SqliteSynchronous::Normal)
```
jbearer, line 240: I'm wondering about this. Based on [the docs](https://www.sqlite.org/pragma.html#pragma_synchronous), this is almost always safe with WAL mode, _however_, the host restarting at just the wrong time could very rarely cause us to lose committed data. Since we base our votes on data being successfully committed, this weakens the guarantee that, e.g., at least one honest node has the leaf corresponding to each QC. Since AWS does have the propensity to restart services without warning, this worries me a bit.

It is probably an extremely low-probability failure, but if it is not prohibitively expensive to run in `FULL` mode, I think we should just do that and be extra robust

src/data_source/storage/sql.rs
```diff
@@ -1051,6 +1072,11 @@ pub mod testing {
 
     impl TestMerkleTreeMigration {
         fn create(name: &str) -> String {
+            #[cfg(feature = "embedded-db")]
+            let bitvec_type = "TEXT";
+            #[cfg(not(feature = "embedded-db"))]
+            let bitvec_type = "BIT(8)";
```
jbearer, line 1078: Might be nice to expose some of these publicly (maybe as constants in the `db` module?). I noticed we also have max function being either `MAX` or `GREATEST` in the sequencer. If we export all these little differences from here perhaps we can keep the `#[cfg]` gating limited to this crate

migrations/postgres/V200__rename_transaction_table.sql
jbearer, line 1: We have already deployed with the aggregates table as version 200, and `num_transactions` column as version 300 is coming. I think after merging in main this will need to be set to 400

migrations/sqlite/V10__init_schema.sql
jbearer, line 1: Any reason not to stick to the same multiple-of-100 convention for sqlite?

migrations/sqlite/V10__init_schema.sql
```diff
@@ -0,0 +1,75 @@
...
+);
+
+CREATE INDEX header_timestamp_idx ON header (timestamp);
+
+CREATE TABLE payload
+(
+    height BIGINT PRIMARY KEY REFERENCES header (height) ON DELETE CASCADE,
+    size   INTEGER,
+    data   BLOB
+);
```
jbearer, line 24: This will need to be updated with `num_transactions` after a merge with main

jbearer (changes requested):

src/data_source/storage/sql.rs
```diff
@@ -545,10 +661,31 @@ impl PruneStorage for SqlStorage {
...
+                name,
+                (page_count * page_size) AS total_bytes
+            FROM 
+                pragma_page_size
+            JOIN 
+                pragma_page_count ON 1 = 1
+            JOIN 
+                sqlite_master ON type = 'table'
+            WHERE 
+                name NOT LIKE 'sqlite_%'
```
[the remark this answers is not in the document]
imabdulbasit: actually yes, this is totally wrong. 

pragma_page_count * page size = file size

migrations/sqlite/V10__init_schema.sql
imabdulbasit: we started as multiples of 10 for postgres so I keep it as it is. I'll change it

src/data_source/storage/sql.rs
imabdulbasit: exported it under syntax_helpers() module in db.rs

jbearer (approved):

src/data_source/storage/sql.rs
```diff
@@ -587,6 +587,8 @@ impl SqlStorage {
                 .await?;
         }
 
+        conn.close().await?;
```
jbearer, line 590: Doesn't this happen automatically when `conn` is dropped?

imabdulbasit: I was trying the locking_mode to exclusive and db was stuck after migrations. closing the connection explicitly fixed it

PR #729, opened 2024-11-07
Weekly release

rob-maron:


src/fetching/provider/testing.rs
```diff
@@ -54,7 +55,7 @@ impl<P> TestProvider<P> {
     pub async fn block(&self) {
         let mut unblock = self.unblock.write().await;
         if unblock.is_none() {
-            *unblock = Some(channel().0);
+            *unblock = Some(broadcast::channel(1000).0);
```
rob-maron, line 58: Unbounded -> `1000` capacity, @jbearer does this matter?

rob-maron: and @imabdulbasit

src/data_source/notifier.rs
```diff
@@ -160,18 +162,15 @@ pub struct Notifier<T> {
     // concurrent subscriptions go through the high-throughput multi-producer stream implementation,
     // and do not contend for the lock on `FilterSender::subscribers`.
     #[derivative(Debug = "ignore")]
-    pending: UnboundedReceiver<Subscriber<T>>,
-    #[derivative(Debug = "ignore")]
-    subscribe: UnboundedSender<Subscriber<T>>,
```
jbearer, line 165: The point of all this is to avoid blocking on an exclusive lock when we subscribe. If we're taking a lock anyways, we can do something much simpler. But I feel like an optimized mpsc channel would still probably be better here.

src/fetching/provider/testing.rs
jbearer: This should be fine, it's only used for tests anyways

src/task.rs
```diff
@@ -125,7 +126,7 @@ impl<T: Send + 'static> Drop for Task<T> {
             // exit quickly: as soon as the background task has been successfully cancelled.
             spawn(async move {
                 tracing::info!(name = inner.name, "cancelling task");
-                inner.handle.cancel().await;
+                inner.handle.abort();
```
jbearer, line 129: Since tokio task cancellation is synchronous, we can get rid of this `spawn` and just call `abort` directly in the drop handler

src/testing.rs
```diff
@@ -20,12 +19,11 @@ pub mod mocks;
 pub async fn sleep(dur: Duration) {
     // For some reason, `async_std::task::sleep` doesn't work on the GitHub Windows runners (it
     // hangs forever). `spin_sleep::sleep` works fine.
```
jbearer, line 21: Comment is obsolete

src/data_source/notifier.rs
rob-maron: We can get all results but it would have to be up to a certain limit. Is there one you would recommend?

src/task.rs
```diff
@@ -124,11 +124,9 @@ impl<T: Send + 'static> Drop for Task<T> {
             // `drop` runs in a synchronous context, so this is the best we can do. In effect we are
             // replacing a long-lived background task with a short-lived detached task which will
             // exit quickly: as soon as the background task has been successfully cancelled.
```
jbearer, line 126: This comment is obsolete now

src/data_source/notifier.rs
rob-maron: Okay - I changed it back to channels

PR #730, opened 2024-11-08
`Hotshot/rc-0.5.81`

rob-maron:


Cargo.toml
```diff
@@ -100,26 +90,33 @@ typenum = "1"
...
+] }
+tracing-subscriber = "0.3"
 
 # Dependencies enabled by feature "file-system-data-source".
 atomic_store = { git = "https://github.com/EspressoSystems/atomicstore.git", tag = "0.1.4", optional = true }
 
 # Dependencies enabled by feature "sql-data-source".
 include_dir = { version = "0.7", optional = true }
-log = { version = "0.4", optional = true }
+log = { version = "0.4" }
```
jbearer, line 106: If this is required now can you move it up (out of the section labeled "Dependencies enabled by feature "sql-data-source".")

examples/simple-server.rs
```diff
@@ -103,10 +102,12 @@ async fn init_data_source(db: &Db) -> DataSource {
...
 }
 
-#[async_std::main]
+#[tokio::main]
 async fn main() -> Result<(), Error> {
-    setup_logging();
-    setup_backtrace();
+    // Initialize logging
+    let _ = tracing_subscriber::fmt()
+        .with_env_filter(EnvFilter::from_default_env())
```
jbearer, line 109: Hmm so we no longer get all the good stuff with like `RUST_LOG_FORMAT={json,full,pretty}` and ANSI colors? Doesn't really matter for this little example but when it comes time to do the sequencer we will want that stuff

src/fetching/provider/testing.rs
```diff
@@ -14,15 +14,16 @@
 
 use super::Provider;
 use crate::fetching::Request;
-use async_compatibility_layer::async_primitives::broadcast::{channel, BroadcastSender};
-use async_std::sync::{Arc, RwLock};
+// use async_compatibility_layer::async_primitives::broadcast::{channel, BroadcastSender};
```
jbearer, line 17: Please delete commented out code before merging

examples/simple-server.rs
rob-maron: I moved this to the `HotShot` crate but I didn't use it for this example because it would need to pull in HS

Cargo.toml
rob-maron: Sure

PR #734, opened 2024-11-14
Remove unused Default bound on InstanceState

sveitser:
In the sequencer repo we don't have a proper `Default` impl but one that's gated behind the testing feature because it wouldn't actually work.

Working on
https://github.com/EspressoSystems/espresso-sequencer/pull/2283 it seems that this is one of the pain points for removing the spread of the testing features in the sequencer crates.

It looks like we don't need the trait bound anymore in the query service.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,9 +6,6 @@ on:
       - main
       - release-*
   pull_request:
-    branches:
-      - main
-      - release-*
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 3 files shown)


sveitser:
We previously remove the bound and then reverted the commit in  https://github.com/EspressoSystems/hotshot-query-service/commit/d61b41e19df80dbce951e262e4c3bd6a7888cff9 

@imabdulbasit @jbearer do you know why we did that?

imabdulbasit (approved):

PR #735, opened 2024-11-15
Improve efficiency of aggregated statistics

jbearer:
We have some statistics that are derived from the entire block history, namely transaction count and payload size. Previously, these were computed on the fly each request, which requires an expensive full table scan and, in the case of payload size, a large sum. We saw this causing performance problems on decaf and mainnet, particularly the block explorer, which uses these statistics heavily

### This PR:

Adds a new table `aggregate`, which stores these cumulative values _at each block height_. This table is kept up to date by a background task scanning the block stream. Now, looking up the values of these statistics just requires reading a single row in this table. This approach has numerous benefits in addition to a massive performance improvement:
* we can now easily look up the values of these statistics at any historical block height, or for any historical _range_ of blocks
* instead of returning inaccurate counts when data is missing, we will simply return that we don't have the counts yet for the requested block height, since the aggregate table won't be populated for that row
* requests that explicitly specify a range or upper bound are easily cachable

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Format Check
         run: cargo fmt -- --check
 
-      # Run Clippy on all targets. The lint workflow doesn't run Clippy on testgs, because the tests
+      # Run Clippy on all targets. The lint workflow doesn't run Clippy on tests, because the tests
       # don't compile with all combinations of features.
       - name: Clippy
         run: cargo clippy --workspace --all-features --all-targets -- -D warnings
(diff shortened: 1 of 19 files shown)


.github/workflows/build.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       - name: Format Check
         run: cargo fmt -- --check
 
-      # Run Clippy on all targets. The lint workflow doesn't run Clippy on tests, because the tests
+      # Run Clippy on all targets. The lint workflow doesn't run Clippy on testgs, because the tests
```
zacshowa, line 37: might be a typo here?

zacshowa (approved):
The queries seem correct to me. I think there might be a typo. If the tests pass I think this looks good!

.github/workflows/build.yml
jbearer: whoops

PR #736, opened 2024-11-15
Fix multi-column ORDER BY clauses

jbearer:
Postgres treats `ORDER BY (col1, col2) [DESC]` _very_ differently from `ORDER BY col1 [DESC], col2 [DESC]`. If there is a multi-column index on `(col1, col2)`, only the latter (without parentheses) will use the index. This is causing some extremely slow queries on Decaf and Mainnet, particularly in the block explorer; these are queries which should be almost trivial.

This may overlap a bit with work done in #728 , however this PR is specifically meant to address this issue exhaustively, and to be standalone, so hopefully we can merge this very quickly, as it's a huge performance improvement.

### This PR:
This changes fixes every occurence of these type of queries.


zacshowa (approved):
LGTM

PR #737, opened 2024-11-17
Avoid unnecessary loading of full payloads and VID common

jbearer:
Both the scanner and aggregator tasks wait for full blocks to be available, but neither of them actually need the full block data. The aggregator only needs a summary, and the scanner only needs to know that the block exists. It is quite wasteful to load the whole payload or VID common, which might be quite large, especially in the scanner where we really don't care about data we already have.

### This PR:
* Adds new types `PayloadMetadata` and `VidCommonMetadata`, which are like their friends `PayloadQueryData` and `VidCommonQueryData`, but without the large data blobs
* Implement fetching traits for these types which do not load the large data blob when the type is available in the local DB, but _do_ trigger a fetch of the whole object when it is missing from the local DB
* Add a column `payload.num_transactions`, so that `PayloadMetadata` can have this field, so that it is sufficient for what the aggregator task needs to do
* Modifies the aggregator and scanner tasks to use these metadata objects instead of the full objects
* Modifies the aggregator task to process blocks in batches when it is catching up, which should speed up and decrease DB load during sync

### This PR does not:
Apply the new metadata types to some explorer queries which could be substantially optimized by using them. This should be done in followup work.

### DIFF
--- a/migrations/V300__transactions_count.sql
+++ b/migrations/V300__transactions_count.sql
@@ -1,0 +1,11 @@
+ALTER TABLE payload
+    ADD COLUMN num_transactions INTEGER;
+
+-- Initialize the `num_transactions` column by counting transactions for each
+-- existing payload.
+UPDATE payload AS p
+    SET (num_transactions) =
+        (SELECT count(*) FROM transaction AS t where t.block_height = p.height)
+    -- Don't set `num_transactions` (leave it NULL) for payloads we don't have
+    -- yet.
+    WHERE p.data IS NOT NULL;
--- a/src/data_source/storage/sql/queries/availability.rs
+++ b/src/data_source/storage/sql/queries/availability.rs
@@ -14,16 +14,17 @@
 
 use super::{
     super::transaction::{Transaction, TransactionMode},
-    QueryBuilder, BLOCK_COLUMNS, LEAF_COLUMNS, PAYLOAD_COLUMNS, VID_COMMON_COLUMNS,
+    QueryBuilder, BLOCK_COLUMNS, LEAF_COLUMNS, PAYLOAD_COLUMNS, PAYLOAD_METADATA_COLUMNS,
+    VID_COMMON_COLUMNS,
 };
 use crate::{
     availability::{
         BlockId, BlockQueryData, LeafId, LeafQueryData, PayloadQueryData, QueryableHeader,
         QueryablePayload, TransactionHash, TransactionQueryData, VidCommonQueryData,
     },
-    data_source::storage::AvailabilityStorage,
+    data_source::storage::{AvailabilityStorage, PayloadMetadata},
     types::HeightIndexed,
-    ErrorSnafu, Header, Payload, QueryError, QueryResult,
+    ErrorSnafu, Header, MissingSnafu, Payload, QueryError, QueryResult,
 };
 use async_trait::async_trait;
 use futures::stream::{StreamExt, TryStreamExt};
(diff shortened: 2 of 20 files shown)


src/data_source/storage/sql/queries/availability.rs
```diff
@@ -96,6 +97,31 @@ where
...
+    ) -> QueryResult<PayloadMetadata<Types>> {
+        let mut query = QueryBuilder::default();
+        let where_clause = query.header_where_clause(id)?;
+        // ORDER BY h.height ASC ensures that if there are duplicate blocks (this can happen when
+        // selecting by payload ID, as payloads are not unique), we return the first one.
+        let sql = format!(
+            "SELECT {PAYLOAD_METADATA_COLUMNS}
+              FROM header AS h
+              JOIN payload AS p ON h.height = p.height
+              WHERE {where_clause} AND p.num_transactions != NULL
```
imabdulbasit, line 112:
```suggestion
              WHERE {where_clause} AND p.num_transactions IS NOT NULL
```
I think you can not compare NULL with equality operators in postgres

src/data_source/storage/sql/queries/availability.rs
```diff
@@ -117,6 +143,27 @@ where
...
+        let mut query = QueryBuilder::default();
+        let where_clause = query.header_where_clause(id)?;
+        // ORDER BY h.height ASC ensures that if there are duplicate blocks (this can happen when
+        // selecting by payload ID, as payloads are not unique), we return the first one.
+        let sql = format!(
+            "SELECT {VID_COMMON_METADATA_COLUMNS}
+              FROM header AS h
+              JOIN vid AS v ON h.height = v.height
+              WHERE {where_clause}
+              ORDER BY h.height ASC
```
imabdulbasit, line 159: why don't we select the most recent one? i.e order by DESC

migrations/V300__transactions_count.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TABLE payload
+    ADD COLUMN num_transactions INTEGER;
+
+-- Initialize the `num_transactions` column by counting transactions for each
+-- existing payload.
+UPDATE payload AS p
```
imabdulbasit, line 6: This seems like an expensive query, and we have over 1 million transactions on the mainnet. I'm not sure how long this will take.
We can also do a temporary table to store transaction counts and then update?

src/data_source/storage/sql/queries/availability.rs
jbearer: The idea is that the response shouldn't change over time. So if you once queried and got an old payload, and then a new one comes with the same hash, you query again you get the same, old one again. There are still plenty of ways to get that new payload if you want it, you just have to specify height, or block hash.

Tbh it doesn't really matter that much, but we should be consistent with the existing behavior unless we have a good reason to deviate from it

imabdulbasit: sounds good!

migrations/V300__transactions_count.sql
jbearer: I think it will be a bit slow but we've run slow migrations before (ie dropping the leaf payloads). It just means the node takes a while to start up initially, but since migrations run before anything else, it doesn't interfere with the performance of the node after it's started.

For reference I tried the following read-only equivalent on Decaf (800k blocks, 900k transactions)
```sql
explain analyze select p.height, (select count(*) from transaction as t where t.block_height = p.height) as num_transactions from payload as p;
                                                                      QUERY PLAN                                                                      
------------------------------------------------------------------------------------------------------------------------------------------------------
 Index Only Scan using payload_pkey on payload p  (cost=0.42..7055578.39 rows=804740 width=16) (actual time=1.453..2877.085 rows=816574 loops=1)
   Heap Fetches: 99568
   SubPlan 1
     ->  Aggregate  (cost=8.73..8.74 rows=1 width=8) (actual time=0.003..0.003 rows=1 loops=816574)
           ->  Index Only Scan using transaction_pk on transaction t  (cost=0.55..8.71 rows=9 width=0) (actual time=0.002..0.002 rows=1 loops=816574)
                 Index Cond: (block_height = p.height)
                 Heap Fetches: 7913
 Planning Time: 3.260 ms
 Execution Time: 2921.088 ms
(9 rows)

```

Expensive for sure but not unreasonable. Notice it only looks at indexes. The update will of course be much slower since writing is slower than reading, but even if it's 100 times slower, the migration would complete in about 5 minutes.

Can you say more about why a temporary table would speed things up?

jbearer: Lol I just ran the actual migration on decaf query-db0 (wrapped in begin/rollback). It took 11 seconds. So I think we are good

imabdulbasit: Okay, perfect! I just wasn’t sure how long it would take. 
I was thinking of creating a new temp table with a GROUP BY on transactions.height, then creating an index on height in that temp table, and finally running the migration with a JOIN on height.

imabdulbasit: 11s wow!

imabdulbasit (approved):

migrations/V300__transactions_count.sql
imabdulbasit: Basically, I thought it would run a count query for n blocks, which I expected to be slower than  a GROUP BY and  join on update, but now I’m not so sure, haha.

PR #739, opened 2024-11-18
Improve robustness of fetches during database instability

jbearer:
We have observed on Mainnet that when the database is having errors (such as performance issues leading to timeouts acquiring a connection), some fetches never return, because we erroneously believe an object is missing and wait for, when really it was just temporarily inaccessible and no one is fetching it.

### This PR:

* Makes low-level fetching-related functions fallible, so we can more easily distinguish between an object being missing vs temporarily inaccessilbe
* Retries fetch initialization until we successfully trigger the fetch

The guarantee that this change upholds is that every fetch will eventually resolve as long as the requested object is either fetchable or exists in the database already.

See commit messages, comments and new tests for additional details.

### DIFF
--- a/src/data_source/fetching.rs
+++ b/src/data_source/fetching.rs
@@ -146,7 +146,7 @@ pub struct Builder<Types, S, P> {
     storage: S,
     provider: P,
     backoff: ExponentialBackoffBuilder,
-    rate_limit: Option<usize>,
+    rate_limit: usize,
     range_chunk_size: usize,
     minor_scan_interval: Duration,
     major_scan_interval: usize,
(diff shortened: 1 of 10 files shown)


src/data_source/fetching.rs
```diff
@@ -911,128 +1081,58 @@ where
...
         // and we do want to log them.
         let ts = ts.into_iter().filter_map(ResultExt::ok_or_trace);
+
         // Kick off a fetch for each missing object.
-        let mut fetches = Vec::with_capacity(chunk.len());
+        let mut results = Vec::with_capacity(chunk.len());
         for t in ts {
             // Fetch missing objects that should come before `t`.
-            while chunk.start + fetches.len() < t.height() as usize {
+            while chunk.start + results.len() < t.height() as usize {
```
imabdulbasit, line 1089: is this while loop because load_range() returns the truncated list? wouldn't this result in lot of unnecessary fetches?

imabdulbasit: nvm  the loop steps ahead because of results.len()

imabdulbasit (approved):

PR #741, opened 2024-11-22
separate transactions when updating aggregates

imabdulbasit:
I am seeing a lot of database locking errors for sqlite during the update of aggregates. I tried separating the read transaction for the aggregate count from the write transaction, and the locking errors are gone. However, I am not sure why this is happening.

### DIFF
--- a/src/data_source/fetching.rs
+++ b/src/data_source/fetching.rs
@@ -1360,24 +1360,30 @@ where
     #[tracing::instrument(skip_all)]
     async fn aggregate(self: Arc<Self>, chunk_size: usize, metrics: AggregatorMetrics) {
         loop {
-            let start = loop {
+            let prev_aggregate = loop {
                 let mut tx = match self.read().await {
                     Ok(tx) => tx,
                     Err(err) => {
-                        tracing::error!("unable to start aggregator: {err:#}");
+                        tracing::error!("unable to open read tx: {err:#}");
                         sleep(Duration::from_secs(5)).await;
                         continue;
                     }
                 };
-                match tx.aggregates_height().await {
-                    Ok(height) => break height,
+                match tx.load_prev_aggregate().await {
+                    Ok(agg) => break agg,
                     Err(err) => {
-                        tracing::error!("unable to load aggregator height: {err:#}");
+                        tracing::error!("unable to load previous aggregate: {err:#}");
                         sleep(Duration::from_secs(5)).await;
                         continue;
                     }
-                };
+                }
             };
+
+            let (start, mut prev_aggregate) = match prev_aggregate {
+                Some(aggregate) => (aggregate.height as usize + 1, aggregate),
+                None => (0, Aggregate::default()),
+            };
+
             tracing::info!(start, "starting aggregator");
             metrics.height.set(start);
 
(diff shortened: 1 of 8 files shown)


src/data_source/storage/sql.rs
```diff
@@ -922,6 +922,7 @@ pub mod testing {
                 let path = format!("sqlite:{db_path}");
                 sqlx::sqlite::SqliteConnectOptions::from_str(&path)
                     .expect("invalid db path")
+                    .locking_mode(sqlx::sqlite::SqliteLockingMode::Exclusive)
```
jbearer, line 925: This seems...not great. IIUC this means the first time a connection locks the database for writing, it will never release that lock. That means any other connection will be unable to access the database until the first is closed, which means we get no parallelism from the connection pool. Would be good to understand if we really need this and why

src/data_source/storage/sql/queries/node.rs
```diff
@@ -286,28 +286,34 @@ impl<Mode: TransactionMode> AggregatesStorage for Transaction<Mode> {
...
+        })
+    }
 }
 
 impl<Types: NodeType> UpdateAggregatesStorage<Types> for Transaction<Write> {
-    async fn update_aggregates(&mut self, blocks: &[PayloadMetadata<Types>]) -> anyhow::Result<()> {
-        // Get the cumulative statistics up to the block before this chunk.
+    async fn update_aggregates(
+        &mut self,
+        aggregate: Aggregate,
```
jbearer, line 309: Nit: `prev` or something is probably a better name for what this is

src/data_source/fetching.rs
```diff
@@ -1401,8 +1402,16 @@ where
...
+                        let aggregate = if start == 0 {
+                            Aggregate::default()
+                        } else {
+                            let mut tx = self.read().await.context("opening transaction")?;
+                            tx.aggregate((start - 1).try_into()?).await?
+                        };
+
                         let mut tx = self.write().await.context("opening transaction")?;
-                        tx.update_aggregates(&chunk).await?;
+                        tx.update_aggregates(aggregate, &chunk).await?;
```
jbearer, line 1414: Since we're now passing in the previous aggregates to each iteration, we have the opportunity to do something kinda cool here...we could have `update_aggregates` also return the updated aggregates, and then we can just keep `aggregate` in a local variable here, updating each iteration of the loop. We would only need to read it from the database once before the first iteration.

This would actually save one database interaction every iteration of the loop which is pretty good. Similar to how the state update loop works in the confirmation layer

jbearer (approved):
Not sure how or why this helps either, but it looks correct.

Also see comment about a nice optimization this unlocks

src/data_source/storage/sql/queries/node.rs
```diff
@@ -340,9 +341,18 @@ impl<Types: NodeType> UpdateAggregatesStorage<Types> for Transaction<Write> {
             "aggregate",
             ["height", "num_transactions", "payload_size"],
             ["height"],
-            rows,
+            rows.clone(),
```
jbearer, line 344: Nit: it is a bit wasteful to clone all the rows here just to grab the last row below. Might make more sense to pull out the updated aggregate first. Doesn't matter much though as the rows are usually small. This is not a blocker

src/data_source/storage/sql/queries/node.rs
```diff
@@ -340,9 +341,18 @@ impl<Types: NodeType> UpdateAggregatesStorage<Types> for Transaction<Write> {
...
         )
-        .await
+        .await?;
+
+        let (height, num_transactions, payload_size) =
+            rows.last().ok_or_else(|| anyhow!("no row"))?;
```
jbearer, line 349: Nit: I think `.ok_or_else(|| anyhow!(whatever))` is more simply written `.context(whatever)`

jbearer (approved):
LGTM!

src/data_source/storage/sql/queries/node.rs
imabdulbasit: yeah, this had conflict with snafu context

PR #742, opened 2024-11-27
Adjust Explorer Query and Error Handling

Ayiga:
### This PR:

Adjusts some error handling to have the intended error encoding representation.

Adjust the way Explorer queries are handled.  Due to the reference of a local variable within the query strings, they are unable to be static.  The usage of a lazy static library allows for the query strings to be statically set.  They were being constructed in a manner that may have allowed for dependency injection if care was not taken.  They are now set separately, and their lifetimes are external to the queries themselves, allowing for more consistent query handling.

Adjusts the histogram data fetched, so that we don't miss a histogram data point due to missing the prior data point in the results.

### DIFF
--- a/src/data_source/storage/sql/queries/explorer.rs
+++ b/src/data_source/storage/sql/queries/explorer.rs
@@ -39,7 +39,7 @@ use futures::stream::{self, StreamExt, TryStreamExt};
 use hotshot_types::traits::node_implementation::NodeType;
 use itertools::Itertools;
 use sqlx::{types::Json, FromRow, Row};
-use std::num::NonZeroUsize;
+use std::{collections::VecDeque, num::NonZeroUsize};
 use tagged_base64::{Tagged, TaggedBase64};
 
 impl From<sqlx::Error> for GetExplorerSummaryError {
@@ -250,6 +250,18 @@ lazy_static::lazy_static! {
         )
     };
 }
+
+/// [EXPLORER_SUMMARY_HISTOGRAM_NUM_ENTRIES] is the number of entries we want
+/// to return in our histogram summary.
+const EXPLORER_SUMMARY_HISTOGRAM_NUM_ENTRIES: usize = 50;
+
+/// [EXPLORER_SUMMARY_NUM_BLOCKS] is the number of blocks we want to return in
+/// our explorer summary.
+const EXPLORER_SUMMARY_NUM_BLOCKS: usize = 10;
+
+/// [EXPLORER_SUMMARY_NUM_TRANSACTIONS] is the number of transactions we want
+/// to return in our explorer summary.
+const EXPLORER_SUMMARY_NUM_TRANSACTIONS: usize = 10;
 
 #[async_trait]
 impl<Mode, Types> ExplorerStorage<Types> for Transaction<Mode>
@@ -471,13 +483,14 @@ where
                 JOIN payload AS p ON
                     p.height = h.height
                 WHERE
-                    h.height IN (SELECT height FROM header ORDER BY height DESC LIMIT 50)
+                    h.height IN (SELECT height FROM header ORDER BY height DESC LIMIT $1)
                 ORDER BY h.height
                 ",
             )
+            .bind((EXPLORER_SUMMARY_HISTOGRAM_NUM_ENTRIES + 1) as i64)
             .fetch(self.as_mut());
 
-            let histograms: Result<ExplorerHistograms, sqlx::Error> = histogram_query_result
+            let mut histograms: ExplorerHistograms = histogram_query_result
                 .map(|row_stream| {
                     row_stream.map(|row| {
                         let height: i64 = row.try_get("height")?;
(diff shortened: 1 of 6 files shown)


src/data_source/storage/sql/queries/explorer.rs
```diff
@@ -105,6 +105,164 @@ where
     }
 }
 
+lazy_static::lazy_static! {
+    static ref GET_BLOCK_SUMMARIES_QUERY_FOR_LATEST: String = {
```
imabdulbasit, line 109: Are these static variables defined just for readability? I see these queries being used only once.

imabdulbasit: and do we have tests for all these queries?

src/data_source/storage/sql/queries/explorer.rs
```diff
@@ -105,6 +105,164 @@ where
...
+        )
+    };
+
+    static ref GET_BLOCK_DETAIL_QUERY_FOR_HASH: String = {
+        format!(
+            "SELECT {BLOCK_COLUMNS}
+                FROM header AS h
+                JOIN payload AS p ON h.height = p.height
+                WHERE h.hash = $1
+                ORDER BY h.height DESC
```
imabdulbasit, line 173: same here. 
 The block hash is unique so I don't understand the purpose of ORDER BY here. Is this because of forcing the query planner to use index search?

src/data_source/storage/sql/queries/explorer.rs
```diff
@@ -105,6 +105,164 @@ where
...
+        )
+    };
+
+    static ref GET_BLOCK_DETAIL_QUERY_FOR_HEIGHT: String = {
+        format!(
+            "SELECT {BLOCK_COLUMNS}
+                FROM header AS h
+                JOIN payload AS p ON h.height = p.height
+                WHERE h.height = $1
+                ORDER BY h.height DESC
```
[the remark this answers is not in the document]
Ayiga: These were just for consistency of the various queries based on what they were querying.  To be clear, these were already here, and were not added with this change, simply relocated.  But you are correct in that they are unnecessary.

src/data_source/storage/sql/queries/explorer.rs
Ayiga: They are primarily being defined as a static reference for a couple of reasons and advantages.

- To ensure that they are indeed a static, unchanging string.

Query strings **should** be largely unchanging.  Otherwise you open yourself up to the potential for SQL injection attacks.  Consider what was being done before:

```rust
  BlockIdentifier::Hash(hash) => format!(
    "SELECT {BLOCK_COLUMNS}
        FROM header AS h
        WHERE h.hash = {}
        ORDER BY h.height DESC 
        LIMIT 1",
    query.bind(hash.to_string())?,
  )
```

 At first glance this seems to be roughly the same thing.  But, taking a closer look, the entire query string is in a `format` macro with a variable replacement / injection being provided by the result of the `query.bind`.  I don't believe this is the intended or correct behavior.  Additionally, attempts to split this up resulted in a difficult to resolve lifetime issue.

- This allows the query strings to be defined once, and then referenced multiple times.

This cleans up a lot of lifetime issues, as the strings are static, and outlive their async blocks.

- It prevents the string from having to be re-computed / formatted for every invocation of the query, saving memory.

- The static unchanging string also allows for `sqlx`'s query caching to be guaranteed to take advantage of its internal prepared statements for re-use.

Ayiga: During the refactor of this commit: 8f64921843d4cb27ad0e88b86a789869c369975a, the separation of the query string definition, and the query parameters were ultimately modified, leading to the potential for an SQL injection, as we are directly inserting a string coming from the end-user.

Ayiga: These queries do get tested indirectly at this symbol:
`explorer::test::validate`.

src/data_source/storage/sql/queries/explorer.rs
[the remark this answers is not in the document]
imabdulbasit: I remember we used LIMIT 1 intentionally to force the query planner to use an index, and I am wondering if that is also the case for ORDER BY. I would suggest keeping the ORDER BY for now and testing the query planner on a local database.

src/data_source/storage/sql/queries/explorer.rs
imabdulbasit: The bind() method of the QueryBuilder returns `self.arguments.len()`  so basically, it provides a placeholder for the query and stores the arguments in its struct.

Ayiga: Oh I see.  I made an assumption as to what this was returning because I couldn't think of any other reason why the result of `bind` would return a `Result<String>`.  So this currently returns a placeholder representation to put into query strings.

I do think that actually utilizing the `bind` call in this way with the `format!` macro is best avoided.

The result of the `bind` function could potentially be modified in the future without changing the API to return a different `String` value in the future which would not cause a breaking change in terms of the API.  I do think that this is unlikely, but the possibility for it to happen is non-zero.

Additionally, since there is a `format!` macro, if in the future another parameter is added to these queries for whatever reason, we leave the possibility of a developer just adding it to the query string via a `format` parameter instead of through the `QueryBuilder` itself.  While we could argue that this would surely be caught in review, I think it's probably best to avoid that possibility from the get go.

imabdulbasit: I agree that this bind() is confusing and should be changed.  I am thinking we store the string with placeholders inside the QueryBuilder and bind() can return the QueryBuilder or it might just take &mut self

imabdulbasit: https://github.com/EspressoSystems/hotshot-query-service/issues/763

imabdulbasit (approved):

PR #743, opened 2024-11-27
[WEEKLY RELEASE] Bump HotShot

QuentinI:
Closes #<ISSUE_NUMBER>


Cargo.toml
```diff
@@ -12,7 +12,7 @@
 
 [package]
 name = "hotshot-query-service"
-version = "0.1.62"
+version = "0.1.63"
```
imabdulbasit, line 15: we haven't updated this in a while. The version should be 0.1.75

src/availability/query_data.rs
```diff
@@ -190,8 +190,8 @@ pub trait QueryablePayload<Types: NodeType>: traits::BlockPayload<Types> {
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 #[serde(bound = "")]
 pub struct LeafQueryData<Types: NodeType> {
-    pub(crate) leaf: Leaf<Types>,
-    pub(crate) qc: QuorumCertificate<Types>,
+    pub(crate) leaf: Leaf2<Types>,
```
imabdulbasit, line 193: hmm I am not too sure about this change. We should be inserting Leaf1 for now or migrate the exisitng leaves to Leaf2 and start inserting leaf2

imabdulbasit: but in the sequencer, we are inserting V1 of everything so I think we want to keep it as Leaf1?

imabdulbasit (approved):

Issue #724, opened 2024-10-29
add test for explorer transaction summaries endpoint


imabdulbasit:
https://github.com/EspressoSystems/hotshot-query-service/pull/723

imabdulbasit:
tests for https://github.com/EspressoSystems/hotshot-query-service/pull/723 and https://github.com/EspressoSystems/hotshot-query-service/pull/722

PR #749, opened 2024-12-05
feat: added tests for transaction details

0xObsidian:
### This PR:
- Added tests to prevent regression of DESC ordering:
  - Referring to the previously fixed issue where DESC + OFFSET incorrectly skipped recent transactions
- Fixed query syntax compatibility between SQLite and PostgreSQL for LIMIT/OFFSET ordering

### Key places to review:
- `src/data_source/storage/sql/queries/explorer.rs`
- The written test in it
- Introduced syntax compatibility between SQLite and PostgreSQL

### Things tested
- Transaction ordering test to prevent DESC regression: 
  ```bash
  cargo nextest run test_transaction_detail_asc_ordering
  ```
- Verified with both `--all-features` and individual feature flags

### Implementation Details
- Added conditional query formatting based on database back-end
- Introduced test for ordering stays ASC to prevent re-introduction of DESC + OFFSET bug that caused incorrect transaction skipping

Previous Error ( in the introduced test prior additionally addressing query compatibility ):
```
near OFFSET: syntax error
```

### DIFF
--- a/src/availability.rs
+++ b/src/availability.rs
@@ -499,7 +499,7 @@ fn enforce_range_limit(from: usize, until: usize, limit: usize) -> Result<(), Er
 mod test {
     use super::*;
     use crate::{
-        data_source::{storage::no_storage, ExtensibleDataSource},
+        data_source::ExtensibleDataSource,
         status::StatusDataSource,
         task::BackgroundTask,
         testing::{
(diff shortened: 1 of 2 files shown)


0xObsidian:
This PR addresses #724 --- specifically ---> #723

